PR #4, opened 2022-01-11
Fixing letter L in the word

craigp1231:
The info text mentioned that L was present, but I was shown in the cells


hannahcode:
Nice catch, thank you

PR #15, opened 2022-01-16
enhancement: apply prettier for automatic code formatting

HaithamLeo:
[Prettier](https://prettier.io/docs/en/why-prettier.html) reformats your code to improve readability for a human by following some (user-defined) conventions i.e. re-indenting, splitting the code into lines, adding white spaces at required places to make it easier and faster to read...[read more](https://prettier.io/docs/en/why-prettier.html).

 Automatic formatting enables higher code quality, especially when you are collaborating in a team and other people need to look at the code you’ve written.


hannahcode:
Thanks for doing this!

PR #18, opened 2022-01-17
Display WinModal on page load

gbear605:
In the original Wordle, the WinModal appears on page refreshes after you have won on a given day. This implements the same feature.


hannahcode:
Thanks for doing this!

PR #21, opened 2022-01-17
refactor: apply enum to control modal state

HaithamLeo:
- Added `Dashboard` component then import it in `App`
- Applied JavaScript `enum` to control Modal state with one `useState` React hook

The `enum` facility in Java is much more flexible, powerful, readable and useful than `enums` in most other languages so we can applied here to control modal state.


hannahcode:
Thanks for the PR, in an effort to keep this simple I'm going to leave the state logic and file organization as it is.

PR #31, opened 2022-01-18
Add heroku file and button

Bockiii:
Hi,

with this, you can host your own version of your code with one click on Heroku free tier.

Greetings


hannahcode:
Thanks for the PR, to keep this simple, I would prefer people fork this project and deploy however they choose.

Bockiii:
I'm not sure why it adds complexity. It's an app.json that defines the heroku buildpac. Heroku offers a free tier so when a user presses that button, he gets his own instance of wordle without having to fork or whatever. Also, the user can recieve updates when you update the repo, which would add a whole lot of overhead for the user if they want to keep their repo up to date to yours.

It's a button. You can remove the button if you want, but adding the app.json will give the people the option to use it on heroku

PR #34, opened 2022-01-20
Capture not enough letters in guess

rbrignall:
Hi,
I added a simple test for when a user hasn't entered enough letters for a valid guess (previous error simply said "Word not found"). Hope this works for you. :)


hannahcode:
Thanks for this!

PR #41, opened 2022-01-22
Fixes #40 - Prevent typing after winning game

### LINKED ISSUE #40: Prevent typing after winning game
After winning game, typing on the keyboard or on-screen keyboard shouldn't yield a result.

samuel-rey:
Added check for isGameWon before typing character.

Fixes #40


hannahcode:
Nice catch, thank you!

PR #44, opened 2022-01-22
Fix #42 - Add no-select to buttons

### LINKED ISSUE #42: Bug: Double tapping a key selects the label
In Firefox and Chrome on Android (untested elsewhere) double tapping on a key will select the text instead of adding that letter twice. This is especially annoying when hitting the delete key repeatedly.

Probably some event capturing needs to happen? Or maybe can make the text unselectable in CSS?

buddhist-uni:
This simply adds the no select css class to the keyboard and About buttons so that double tapping on them doesn't select their text.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -154,7 +154,7 @@ function App() {
 
       <button
         type="button"
-        className="mx-auto mt-8 flex items-center px-2.5 py-1.5 border border-transparent text-xs font-medium rounded text-indigo-700 bg-indigo-100 hover:bg-indigo-200 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500"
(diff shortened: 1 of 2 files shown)


diondiondion:
A better fix would be to make them actual `<button>` elements.

buddhist-uni:
> actual `<button>` elements

They already are.

diondiondion:
> They already are.

Not the keys (or the icon buttons in the top right).
https://github.com/hannahcode/wordle/blob/3024a19b69a17cd16bdd74426b5dcb5255cab80c/src/components/keyboard/Key.tsx#L34

buddhist-uni:
Okay. Made it a button... But the bug still repros without the no-select.  Seems the default style here doesn't set no-select on buttons?

hannahcode:
I think the buttons might not have no-select on them by default because of tailwind (it normalizes and removes default styling). Anyways, this looks good so I'll go ahead and merge. Thanks all!

PR #49, opened 2022-01-23
Added animation to cell when a new char is inserted

davidmaillo:
The cells that are filled also get highlighted with a black border.


hannahcode:
Looks good, thanks!

davidmaillo:
You're welcome :) Glad you liked it!

PR #50, opened 2022-01-23
Refactor modals to reduce code duplication

jcserv:
Creates a BaseModal component as a standardized modal interface,
for which the other modals can pass in title and children to reduce code duplication.


hannahcode:
Looks great - thanks so much!

PR #62, opened 2022-01-25
histogram width divide by mode, not total

postylem:
Divide by the value that occurs the most frequently, instead of the total games, so the histogram always uses the whole space available, rather than possibly getting smaller and smaller the more games are played.


hannahcode:
Lgtm, thanks!

PR #64, opened 2022-01-26
Fix #52: Put countdown timer and share button in stats modal

### LINKED ISSUE #52: Merge share and statistics modals
The "reference" wordle puts the share button (and a timer) at the bottom of the statistics modal. This is preferable as it gives the user a way to get back the share button after dismissing the modal.

buddhist-uni:
Delete the old "Win" modal, and put the time and share button in the stats modal like so:

new_stats_modal

This PR also makes the following, smaller changes:
- Opens the stats modal on losses as well as wins
- Fixes the share string in the loss case to "X/6" (Thanks @eggplants !)
- Introduces a nascent "constants/strings" file for UI Strings
- Puts the Alert time in a const
- Gives the user a congratulatory Alert on winning the game before the stats modal opens
- Better stats null state (doesn't try to show the histogram if there's no data)
- Disable "onEnter" if the game was lost already
- Make the "You lost" Alert message a little softer :)


eggplants:
When player lose the game, `n/6` should be `X/6`.

hannahcode:
Looks great, thanks!

ahmetbombaci:
> Looks great, thanks!

I am not sure if it is related to this PR but share is totally broken now. It was okay yesterday.

I am not sure when it happened exactly but I was able to share after reloading the page recently. I think it was working smoothly at some point (maybe a week ago??). Starting from today, the page appears briefly after solving the puzzle and everything disappears after a few seconds.

Browser: Google Chrome

eggplants:
@ahmetbombaci Could you open new issue, share your environments (OS, Browser version), and paste screen captured video?

PR #65, opened 2022-01-26
Fix share text (`6/6` -> `X/6` if losing game)

eggplants:
For #64, if a player loses the game, the won time in the share text should be `'X'`.


buddhist-uni:
Incorporated this fix into my PR: https://github.com/hannahcode/wordle/pull/64   I prefer my solution of just passing the game lost flag. Seems more straightforward than digging into the emoji grid.

eggplants:
@buddhist-uni Got it.

buddhist-uni:
Thanks for pointing it out! :)

PR #67, opened 2022-01-26
Loop solutions

postylem:
It might be a good idea to loop the index, as repeating the words would probably be better than causing an out of range error if the list is exhausted (quite easy to do if the wordlist is modified, as might be what's going on in https://github.com/hannahcode/wordle/issues/63)


hannahcode:
Great idea, thanks!

PR #68, opened 2022-01-26
Use local timezone instead of UTC-5

gbear605:
Currently the epoch is based on Saturday, January 1, 2022 0:00:00 UTC-5. This changes it to use the local timezone. It's also cleaner if other forks want to use a different date for their epoch.


gbear605:
This is also helpful once daylight savings time starts - the people who are currently in the "right" timezone of UTC-5 will have it be wrong for them.

hannahcode:
Looks great! Thanks for doing this.

PR #70, opened 2022-01-28
Close #36: Add dark mode toggle

### LINKED ISSUE #36: Add dark mode
It'd be nice if this had a dark mode like the original Wordle.

I'd make a PR for it but I'm not that familiar with Tailwind so am not 100% sure where to start looking.

arobison203:
This PR adds dark mode using Tailwind's `class` strategy, and allows the user to select either mode, storing the preference in `localStorage`. When a user first visits, we use the media query `prefers-color-scheme` to get the user's preferred color mode, and set the initial state accordingly.


eggplants:
You can lint with `npm run fix`

hannahcode:
Awesome, thanks again for this.

PR #73, opened 2022-01-28
Add RefreshIcon to top menu

elahmo:
When app is added to homescreen, sometimes when a new day is opened it is still loaded from memory and it cannot show the new word prompt.

Pull to refresh doesn't work, so force closing the 'app' is the only solution.

This adds a simple button that allows reloading of the page.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { InformationCircleIcon } from '@heroicons/react/outline'
+import { InformationCircleIcon, RefreshIcon } from '@heroicons/react/outline'
 import { ChartBarIcon } from '@heroicons/react/outline'
 import { useState, useEffect } from 'react'
 import { Alert } from './components/alerts/Alert'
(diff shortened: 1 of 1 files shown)


diondiondion:
Maybe it would be better to automatically refresh the page if the app was open for more than a day.

Alternatively if this solution is accepted, it would be best to only show the icon of the web app is used in standalone mode. This can be achieved using the `display-mode` media query: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode.

hannahcode:
Thanks @elahmo and great points @diondiondion. I'm looking into a solution using the page visibility API (though I'm not sure if there would be an issue accessing this in standalone mode). I'll leave this open for now and circle back!

gheja:
I'd suggest a different approach: the game already handles [solutionIndex](https://github.com/cwackerfuss/react-wordle/blob/main/src/lib/words.ts#L87), we could validate this when the game is gets focus and reload the page if it's out-of-date.

PR #75, opened 2022-01-29
Move alert up to not hide guesses

gbear605:
Currently the alert displays partly over the first guess, hiding what the guess was. This can be frustrating if the alert was about that guess, such as the guess not being long enough or not being a word. This moves it up so that it fully covers the top bar and does not cover any of the guesses.


gbear605:
Before the change:
Before Change

After the change:
After Change

hannahcode:
This looks good - the only issue is that we changed the losing alert to persist, which with this change would cover the title and info and stats buttons. Maybe we can add margin to the very top of the app and have the alerts show above?

gbear605:
Good idea! I've moved the app down a bit so that the alerts don't cover up anything

hannahcode:
Great, thanks!

PR #77, opened 2022-01-29
Extract strings to constants

gbear605:
Currently the text of the app is spread throughout the code, which makes translation into other languages difficult. This PR extracts all of the simple strings to the `strings.ts` file and adds instructions to the README about all of the places that need to be updated to translate the app.

One interesting case was the "The word was ..." message, since the "..." will be in different places in different languages and can't just be appended. I decided to make it a function that takes in the solution as a parameter, but an alternate solution would be to split the before and after texts into two separate constants. Another interesting case was the enter and delete text, which I extracted but could equally have left in the keyboard.

The "about" and "info" modals were not extracted, since they have HTML tags throughout that are easier to update in place, especially if you want the translated versions to not be one-to-one.

I also updated some of the meta files, including `manifest.json`, to refer to "Not Wordle" instead of "Wordle Clone" or the create-react-app default text.


hannahcode:
Thanks @gbear605, this is great. I fixed a small conflict to keep this consistent with the change from your other PR (top margin on the app).

PR #84, opened 2022-01-31
feat<idea>: add .idea to .gitignore por favor

vvmspace:
msg me
Telegram: vvmspace

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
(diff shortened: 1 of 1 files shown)


logo4poop:
This should probably be done on the users local gitignore instead of in the repository itself.

vvmspace:
> This should probably be done on the users local gitignore instead of in the repository itself.

As a way, but not for most users.

cwackerfuss (approved):

PR #85, opened 2022-01-31
getIndex() + renamed tomorrow to next

vvmspace:
From https://wordle.uno with ❤️❤️❤️


src/App.tsx
```diff
@@ -31,6 +36,7 @@ import './App.css'
 const ALERT_TIME_MS = 2000
 
 function App() {
+  const index = getIndex()
```
cwackerfuss, line 39: `solutionIndex` is already being exported by `./lib/words` so let's just import that and leave `getIndex` unexported for now.

I think the `getWordOfDay` function should be refactored in the near future so we're evaluating the word from React rather than through vanilla JS initialization.

src/App.tsx
```diff
@@ -151,7 +157,9 @@ function App() {
   return (
     <div className="py-8 max-w-7xl mx-auto sm:px-6 lg:px-8">
       <div className="flex w-80 mx-auto items-center mb-8 mt-12">
-        <h1 className="text-xl grow font-bold dark:text-white">{GAME_TITLE}</h1>
+        <h1 className="text-xl grow font-bold dark:text-white">
+          {GAME_TITLE} #{index}
+        </h1>
```
cwackerfuss, line 162: Nit: lets remove this from the header, the original game doesn't include the index in the title

src/components/modals/StatsModal.tsx
```diff
@@ -59,7 +59,7 @@ export const StatsModal = ({
             <h5>{NEW_WORD_TEXT}</h5>
             <Countdown
               className="text-lg font-medium text-gray-900 dark:text-gray-100"
-              date={tomorrow}
+              date={next}
```
cwackerfuss, line 62: Nit: while we're changing this, can we be even more expressive and say `nextDayInMs`

cwackerfuss (changes requested):
Looks great, thanks for the thoughtful changes. Just a few comments

src/components/modals/StatsModal.tsx
vvmspace: I am using 1 hour delay.

vvmspace: May be nextWordInMs?

src/lib/words.ts
```diff
@@ -1,6 +1,14 @@
 import { WORDS } from '../constants/wordlist'
 import { VALIDGUESSES } from '../constants/validGuesses'
 
+// TODO: move to constants
+// TODO: process.env there?
+export const MS_IN_DAY = 86400000;
+export const DELAY = 1;
+export const INTERVAL = MS_IN_DAY / DELAY;
+export const START_DATE = '01-01-2022';
```
noemartinez, line 9: I've recently started experimenting with this repo and found that it'd be great to have this constant promoted into ```settings.ts``` and advertised in the README's FAQs as a setting that can be easily adjusted. It might be cool to do that as part of this PR. Thanks for the consideration!

vvmspace:
Guys, not working on it now. Can someone fix merge conflicts etc?

PR #89, opened 2022-02-01
Use 1-based solution index when sharing to clipboard

igitur:
On the first day, the internal `solutionIndex` is 0, but when sharing this to the clipboard it should be 1 (according to move people's conventions).

FYI, I used this to do an Afrikaans version. https://spel.wrintiewaar.co.za

Thanks for a great clone.


cwackerfuss (approved):
good catch! thanks :)

cwackerfuss:
Actually on second thought, i'm thinking we should leave this how it is. The index doesn't need to represent which number solution it is, and I think it's easier to debug if the number we see on the share is consistent with the actual solution index.

gheja:
Also, it would break continuity which might bother players who solved all the daily puzzles so far. The corner case is the first one which, well... already passed.

PR #92, opened 2022-02-01
Typo fix in README.md

rezKamal:
The path to Keyboard.tsx is incorrect in the description of how to change the language of Wordle. It's written as "src/lib/components/keyboard" but it's "src/components/keyboard"


gbear605 (approved):

cwackerfuss (approved):
good catch thanks @rezKamal

PR #97, opened 2022-02-02
Adds character animation on completion.

wuori:
This PR adds a sequence of character reveals similar to the official Wordle app. 

Tested in Chrome 97.0.4692.71 (Mac OSX) and iOS Safari (latest).

Note: I don't work in React much so there may be a better way to hook this in, but the supplied version works fine and even reapplies the animations on page reload.

See GIF for example:  https://imgur.com/dagVbNh

Thank you for sharing this project!

image

### DIFF
--- a/src/components/grid/CompletedRow.tsx
+++ b/src/components/grid/CompletedRow.tsx
@@ -1,5 +1,6 @@
 import { getGuessStatuses } from '../../lib/statuses'
 import { Cell } from './Cell'
+import { useEffect } from 'react'
 
 type Props = {
   guess: string
@@ -8,6 +9,20 @@ type Props = {
 export const CompletedRow = ({ guess }: Props) => {
   const statuses = getGuessStatuses(guess)
 
+  useEffect(() => {
+    // Show completed row cell-by-cell
+    const cellShowDelay = 400;
+    const cellsContainer = document.querySelectorAll('.completed:not(.displayed)') ?? [];
+    if(cellsContainer[0]){
+      Array.from(cellsContainer[0].children).forEach((cell, i) => {
+        setTimeout(() => {
+            cell.classList.add('shown');
+        }, cellShowDelay*i);
+      });
+      cellsContainer[0].classList.add('displayed');
+    }
+  });
+  
   return (
     <div className="flex justify-center mb-1">
       {guess.split('').map((letter, i) => (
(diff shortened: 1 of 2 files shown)


src/components/grid/CompletedRow.tsx
```diff
@@ -8,6 +9,20 @@ type Props = {
 export const CompletedRow = ({ guess }: Props) => {
   const statuses = getGuessStatuses(guess)
 
+  useEffect(() => {
+    // Show completed row cell-by-cell
+    const cellShowDelay = 400;
```
cwackerfuss, line 14: Lets speed this up, with the style of animation, it feels painfully slow. Maybe 100 or 150 ms instead?

src/components/grid/CompletedRow.tsx
```diff
@@ -8,6 +9,20 @@ type Props = {
...
+    const cellsContainer = document.querySelectorAll('.completed:not(.displayed)') ?? [];
+    if(cellsContainer[0]){
+      Array.from(cellsContainer[0].children).forEach((cell, i) => {
+        setTimeout(() => {
+            cell.classList.add('shown');
+        }, cellShowDelay*i);
+      });
+      cellsContainer[0].classList.add('displayed');
+    }
+  });
```
cwackerfuss, line 24: It seems like this isn't formatted -- can you please run `npm run fix` and commit? Thank you!

cwackerfuss (changes requested):
Looks good, few comments

diondiondion:
The approach of adding CSS classes directly to the DOM is very un-react-y and prone to break (React won't know about those classes and discard them on re-renders).

IMO this should be implemented using simple CSS animations with different `animation-delay` timings, not in JS.

cwackerfuss:
@diondiondion I was moving too quickly through PRs and totally missed this. Great call out. @wuori does @diondiondion's suggestion make sense to you? 

It will be a little tricky with CSS to get this behavior, but if possible that is the preferred approach. Otherwise a clean JS solution that dynamically adds/removes an animation class to the React child elements (not using dom methods) is also fine

wuori:
@cwackerfuss Yeah @diondiondion is probably right. I'm not a React developer and didn't consider this, was just playing around and thought it would help make the port more like the original. I'll look into it some more and update this PR if/when I find a better solution.

cwackerfuss:
I think this was actually added in another pr. Thank you for trying -- closing this

PR #98, opened 2022-02-02
Updates character styles.

wuori:
Enlarges characters and adds a slight shadow to improve readability and more closely match the official Wordle application.

See attached for example.

image


cwackerfuss (approved):
Surprised tailwind doesn't support text-shadow. Thanks !

PR #99, opened 2022-02-02
Updates alert styling.

wuori:
Gives alert margins on small screens and limits width on larger screens. See attached for example.

This improves the UX in my opinion but feel free to discard! Thanks!

image


oradwell:
I like this. it's a better use of the screen

cwackerfuss:
thanks @wuori -- Going to address this a slightly different way, just opened a PR, feel free to check it out.

PR #102, opened 2022-02-02
keyboard suggestion

vvmspace:


cwackerfuss:
Thanks for taking a stab at DRYing this up. I think with all the folks making forks of this app in other languages and unique formats, I think we should avoid introducing too much abstraction so that we're not making it more difficult for folks to customize the keys however they want. With this change, we'd be more strongly enforcing the key is a string and the backspace/enter keys are in specific positions. Thoughts @vvmspace ?  I'll leave this open for a few days to leave room for discussion.

PR #109, opened 2022-02-03
chore: generic branding

### LINKED ISSUE #86: Removal of Wordle brand from application?
With NYT's purchase of Wordle, they will no doubt file for trademark. I assume that they get legal precedent over anyone with the name Wordle in the app itself.

Since this is the largest clone repository I could find, may I suggest potentially going for a change in branding to keep this project alive!

dcwarwick:
One more :-)

Closes #86

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "wordle",
+  "name": "game",
   "version": "0.1.0",
   "private": true,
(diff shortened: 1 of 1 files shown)


cwackerfuss (approved):

cwackerfuss:
Good catch :D

PR #112, opened 2022-02-04
feat: Add production build stage to the Dockerfile

erikmartino:
Modifications:

Split the dockerfile into stages

1. Refactored the dockerfile and introduced an `node_modules` stage, where install is cached unless package.json or package-lock.json changes
2. Added prod_builder stage that builds the app, it only runs for prod builds
3. Added a prod stage that creates an image with nginx and the compiled app
4. Refactored the old Dockerfile to build on top of the node_modules stage

Results:

You can either build, by default, the dev image or the prod image by adding `--target=prod` to docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,15 @@
-FROM node
-
-COPY . . 
+FROM node AS node_modules
+WORKDIR /app
+COPY package-lock.json package.json ./
 RUN npm install
+COPY . .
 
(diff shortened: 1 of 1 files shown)


cwackerfuss:
Thank you for the contribution @erikmartino - appreciate the notes since I'm a novice with Docker. :)

cwackerfuss (approved):

PR #121, opened 2022-02-06
Added cell animation on row complete.

TheKnowles:
This PR adds per cell X axis rotation on word submission. 

From left to right, each letter cell will rotate on its X axis and reveal its status at the end of the animation.

This is achieved with css animation and a little state keeping on the react side.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -55,6 +55,7 @@ function App() {
       : false
   )
   const [successAlert, setSuccessAlert] = useState('')
+  const [revealing, setRevealing] = useState(false)
   const [guesses, setGuesses] = useState<string[]>(() => {
     const loaded = loadGameStateFromLocalStorage()
     if (loaded?.solution !== solution) {
@@ -138,6 +139,8 @@ function App() {
       }, ALERT_TIME_MS)
     }
 
+    setRevealing(true)
+
     const winningWord = isWinningWord(currentGuess)
 
     if (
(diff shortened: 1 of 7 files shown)


diondiondion:
Cool, this is a great start!

Just two notes:
- The letters rotate along with the background which looks quite glitchy (as they only do a half rotation and then snap into place). Ideally the flip animation would only affect the background with the letters remaining static over the top. This would probably require a refactor to separate the background to its own element or pseudo element.
- When winning the game, the alert pops up before the animation has finished, spoiling the effect a bit. Would be great if it could be delayed by the duration of the animation

TheKnowles:
Thank you for the feedback.

Timer delays have been updated to not spoil the reveal of a success or failure.

Cell now has a child div for the letter content. A second animation plays on this child element to avoid the half rotation/snap back glitch.

TheKnowles:
Noticed I was behind on main. Rebased, squashed, and force pushed.

cwackerfuss:
Thank you @TheKnowles and @diondiondion for reviewing. I'm going to merge but I have an idea for how to make this work for any guess length. There are some clones that change the guess length so it'd be nice to support that. If I can get it to work, I'll post a PR and tag both of you

soumendrak:
@TheKnowles 
Can we please delay the keyboard highlight until the cell animation is complete?

cwackerfuss:
@soumendrak I decided to push the changes I made to @TheKnowles code. It will now delay the keyboard highlight until cell animation is complete

PR #127, opened 2022-02-07
Reflect color reversal into the share sheet as well

teddyh-io:
I noticed that you decided to change back the color scheme to reflect the original game, but forgot to change the share function back as well.


cwackerfuss (approved):
thanks!!

PR #131, opened 2022-02-08
Fixed lost game timings

TheKnowles:
Fixed lost game timings to allow for final row reveal and small delay to stats modal.

After rebasing, I noticed that on game loss the reveal happens immediately in the alert defeating the purpose of the final row reveal. Additionally, the stats model pops up immediately as soon as the alert snackbar fires. A little delay allows for the user to digest the information before the popup happens. This behavior more closely models wordle in that regard.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,6 +25,7 @@ import {
   MAX_CHALLENGES,
   ALERT_TIME_MS,
   REVEAL_TIME_MS,
+  GAME_LOST_INFO_DELAY,
 } from './constants/settings'
 import { isWordInWordList, isWinningWord, solution } from './lib/words'
 import { addStatsForCompletedGame, loadStats } from './lib/stats'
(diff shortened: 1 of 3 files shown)


cwackerfuss:
Good call - thanks @TheKnowles... what do you think about making `GAME_LOST_INFO_DELAY` be based on `MAX_WORD_LENGTH * REVEAL_TIME_MS`? Or if you like a little buffer between reveal and alert, `(MAX_WORD_LENGTH + 1) * REVEAL_TIME_MS`

TheKnowles:
Went ahead and rolled in the fix to have the alert bars not overlay the top part of the word grid.

cwackerfuss:
Great, thanks! I eventually want to think of a better way to handle this alert -- I don't love the top of the page having so much margin

TheKnowles:
Putting it between the top of the grid and the top menu bar may be the most elegant option.

PR #136, opened 2022-02-08
Implemented a way to hide the solution from being directly accessed

abrarhayat:
Implemented a way to hide the solution from being directly accessed from local storage. There is another solution where we save the index in local storage but that is still very easy for small word lists. Let's make everyone work much harder to get the solution! 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "@types/react": "^17.0.38",
     "@types/react-dom": "^17.0.11",
     "classnames": "^2.3.1",
+    "crypto-js": "^4.1.1",
     "react": "^17.0.2",
     "react-countdown": "^2.3.2",
     "react-dom": "^17.0.2",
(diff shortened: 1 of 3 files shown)


gbear605:
Personally I view this sort of thing as folly, since it's almost as easy (or easier?) to open a private browsing window and type in a random word six times. Wordle is basically insecure by fundamental design, and cryptography won't help.

That said, the implementation looks clean and correct, so your code in itself looks good to me.

TheKnowles:
This is a solution for a problem that does not exist. 

If people want to cheat at their daily word game, no amount of programming is going to stop them.

cwackerfuss:
Agreed with @TheKnowles and @gbear605 -- I don't see the need for this, but I'm glad you put the idea out there. I'll close this.

P.S. rather than importing a new module, I think a similar effect could be achieved with [btoa](https://developer.mozilla.org/en-US/docs/Web/API/btoa) and [atob](https://developer.mozilla.org/en-US/docs/Web/API/atob)

PR #138, opened 2022-02-09
Added hard mode

TheKnowles:
Added hard mode. Game mode is togglable and saved to browser local state. Guesses must use all previously discovered letters.

edit: Forgot share. Share now also adds the asterisk to indicate game was completed in hard mode.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,6 +3,8 @@ import {
   ChartBarIcon,
   SunIcon,
   MoonIcon,
+  CakeIcon,
+  AcademicCapIcon,
 } from '@heroicons/react/outline'
 import { useState, useEffect } from 'react'
 import { Alert } from './components/alerts/Alert'
@@ -26,7 +28,12 @@ import {
   ALERT_TIME_MS,
   REVEAL_TIME_MS,
 } from './constants/settings'
-import { isWordInWordList, isWinningWord, solution } from './lib/words'
+import {
+  isWordInWordList,
+  isWinningWord,
+  solution,
+  findFirstUnusedReveal,
+} from './lib/words'
 import { addStatsForCompletedGame, loadStats } from './lib/stats'
 import {
   loadGameStateFromLocalStorage,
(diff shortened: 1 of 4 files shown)


TheKnowles:
I did not realize that Wordle will enforce letter location on correct reveal as well. Code has been updated to handle all correct location enforcement in addition to any present/discovered letters as well.

cwackerfuss (approved):

cwackerfuss:
LGTM!

gheja:
How about an even harder, adversarial mode like [Absurdle](https://qntm.org/files/absurdle/absurdle.html)? 😈

_"Adversarial" means that Absurdle is actively trying to avoid giving you the answer. With each guess, Absurdle reveals as little information as possible, changing the secret word if need be._
More: https://qntm.org/absurdle

It calculates all possible answers (clues) and how many possible words would remain in each case, then chooses the answer that results in the biggest set. It's fun but can be frustrating. (Also, it logs the calculation and decision in the console.)

Just kidding, this is a whole other gameplay and basically a different game.

igitur:
Just a general comment: if new features are added, especially features that digress from the standard Wordle game, I think it's required that some help text (whether in a tooltip or some form) is also required. 

In this case, Wordle does have a hard mode, but I think a tooltip would've been really useful anyway.

igitur:
Sadly, this wasn't accepted. Else it would've been easy to add tooltips via the svg title elements. https://github.com/tailwindlabs/heroicons/issues/111

TheKnowles:
I agree. It is on my radar to move to a settings icon/popup.

TheKnowles:
I've created a settings modal and moved hard mode toggling to there.

https://github.com/cwackerfuss/react-wordle/pull/146

igitur:
For future PRs, where relevant, please remember to include support for internationalization in all future PRs, or at least include instructions in `README.md`. It's a bit embarrassing for us downstream maintainers of alternate language versions when an English phrase slips in.

ericdrgn:
@TheKnowles would there be a way to autofill correct placement of letters in the current row when in hard mode? since you can't put perfectly placed letters anywhere else anyway I think it might make the experience better when in hard mode.

TheKnowles:
@ericdrgn that's an interesting idea. I would have to contrmplate my navel on how to do it. It would be a deviation from the source material, so not sure if that should be rolled in or maintained in a fork. 

I would personally like this feature as it would help with visualizing new guesses but might actually make hard mode easier?

igitur:
As mentioned elsewhere, my preference is for this repo to stay faithful to the original Wordle game.

ericdrgn:
> @ericdrgn that's an interesting idea. I would have to contemplate on how to do it. It would be a deviation from the source material, so not sure if that should be rolled in or maintained in a fork.
> 
> I would personally like this feature as it would help with visualizing new guesses but might actually make hard mode easier?

I don't know that it makes it easier just more intuitive since it just saves some time/effort. But I see your point.

Either way are you interested in at least helping me figure it out? It seems some people around here are incredibly adamant on staying as close to the original as possible. Which is fine I do get it, but not knowing _everything_ about this kind of code myself I am unable to figure out how to do it in my own fork(s) on my own and since this is the source I figured it was the best place to ask.

PR #142, opened 2022-02-09
Added some animations for the action buttons

abrarhayat:
Added some animations for the svg icons on the action buttons to have an overall animated application.


cwackerfuss:
That's great -- love it! Thanks @abrarhayat

PR #154, opened 2022-02-11
Add configurable Google Analytics

mdanka:
If `REACT_APP_GOOGLE_MEASUREMENT_ID` is provided in `.env` AND the app is running in production, the relevant Google Analytics scripts are included in `index.html`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,3 +94,7 @@ _Want to add one to the list? Just make a pull request or [let us know via a com
 - If the language has letters that are not present in English update the keyboard in [src/components/keyboard/Keyboard.tsx](src/components/keyboard/Keyboard.tsx)
 - If the language's letters are made of multiple unicode characters, use a grapheme splitter at various points throughout the app or normalize the input so that all of the letters are made of a single character
 - If the language is written right-to-left, prepend `\u202E` (the unicode right-to-left override character) to the return statement of the inner function in `generateEmojiGrid` in [src/lib/share.ts](src/lib/share.ts)
+
+### How can I add usage tracking?
+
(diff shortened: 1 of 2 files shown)


cwackerfuss (approved):

igitur:
I think this change is something that should've been left to the custom implementers of this repo to decide. I don't think including Google Analytics by default is the correct way to go. Especially in light of rulings that GA violates GDPR. https://iapp.org/news/a/cnil-is-latest-authority-to-rule-google-analytics-violates-gdpr/

igitur:
Just a general comment. I see this repo has a lot of activity with issues and pull requests. But keep in mind that, unlike most other open-source repos, this repo serves only as a base for others to implement. I hope you agree on that. So I think a measured approach to accepting PRs would be best, rather than just blindly accepting anything. Think about how any change affects downstream repos.

mdanka:
It is not included by default.

igitur:
> It is not included by default.

Would you mind explaining what you mean? Because commit 5fc385f0e5d5ac1a62d7f1c9d2f9cbeafd7569f1 has been merged on the `main` branch. To me that means it's included by default.

igitur:
Sorry, I see this line only now:

```
<% if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_GOOGLE_MEASUREMENT_ID != null) { %>
```

Ok, I understand.

igitur:
Just as a side note: including Google Analytics without asking the user's permission is [not GDPR compliant](https://www.cookiebot.com/en/google-analytics-gdpr/).

> Under the EU’s GDPR, you are required to ask for and obtain the explicit consent to run any kind of cookie or tracker on your website that processes personal data.

So including support for GA, without including support for permission requests is problematic.

PR #157, opened 2022-02-11
Black squares when sharing in dark mode

bluswimmer:
Makes the squares copied to the clipboard when pressing the share button black when playing in dark mode.


cwackerfuss:
Good idea -- can you please run `npm run fix` to address lint failure

Issue #158, opened 2022-02-11
Letter animation should reveal the state of the guess only on the flip side


gheja:
**Is your feature request related to a problem? Please describe.**
The state of the guessed letters can be seen too early during the flipping animation. It would be nice if the states could only be seen when the "other side" of the letter is visible as if it was a physical object with two sides. Like in Wordle.

**Describe the solution you'd like**
Only see the colors when letters are flipped.

**Describe alternatives you've considered**
None really, it is just a cosmetic feature.

**Additional context**
None.

PR #165, opened 2022-02-11
fix(reveal): Change color after reveal

ndench:
It can be quite disorientating for all the keys to turn grey as soon as enter is pressed, and then change color after the reveal animation. This makes them stay the original color and only change to the final colour after the animation.

<details>
<summary>GIF: Light mode</summary>

Peek 2022-02-12 09-12

</details>

<details>
<summary>GIF: Dark mode</summary>

Peek 2022-02-12 09-11

</details>

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -5,11 +5,17 @@
 :root {
   --animation-speed: 1000ms;
   --animation-speed-fast: 500ms;
-  --default-cell-bg-color: theme('colors.slate.400');
+  --default-cell-bg-color: theme('colors.white');
+  --default-cell-border-color: theme('colors.black');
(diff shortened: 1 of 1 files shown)


gheja:
I have a similar PR that makes the color transition happen at the middle of the switch as they were physical object with two sides: https://github.com/cwackerfuss/react-wordle/issues/158 (Currently the transition from base to final starts at the beginning of the flip so it can be seen earlier than it should appear.)

I wanted to make this change too but could not come up a clean way doing it, I see you defined new colors as well. As my changes are smaller I will add them over this PR if you don't mind. (I just have to figure out how to do this :D)

cwackerfuss (approved):
LGTM! Thanks.

PR #168, opened 2022-02-12
Added High Contrast Mode

bluswimmer:
Added a high contrast mode that can be turned on in the settings modal. This changes the squares from yellow and green to cyan and orange, including when you share your game.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -62,6 +62,11 @@ function App() {
       ? true
       : false
   )
+  const [isHighContrastMode, setIsHighConstrastMode] = useState(
+    localStorage.getItem('contrast')
+      ? localStorage.getItem('contrast') === 'high'
+      : false
+  )
   const [successAlert, setSuccessAlert] = useState('')
   const [isRevealing, setIsRevealing] = useState(false)
   const [guesses, setGuesses] = useState<string[]>(() => {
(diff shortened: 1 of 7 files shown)


bluswimmer:
Need to make an additional commit to resolve a visual bug, will report back in a bit...

cwackerfuss (changes requested):
I think it would be more useful to call the localStorage item `highContrast` or something and set its value to either `"true"` or `"1"`. We are not maintaining the "low" value, so lets just delete this if it gets unset.

I know this practice hasn't been used consistently, but lets abstract the use of localStorage into helper fns. There is a file called `localStorage.lib` that we can put these in. I think the React app shouldn't have to do an equality check inline either with this method if you just make the get helper return a boolean.

For example, we can set the `isHighContrastMode` like so:
```
const [isHighContrastMode, setIsHighConstrastMode] = useState(() => getStoredIsHighContrastMode())
```

cwackerfuss (approved):
Looks good to me - one improvement you could make if you'd like to is to call `getStoredIsHighContrastMode` only once and store as a variable in each of the files it's being called several times, but I'm fine also merging this and addressing that later.

cwackerfuss:
And yes -- I do think we should do dark mode / hard mode in the same way. Great point

PR #174, opened 2022-02-13
Letter animation reveals the state of the guess only on the flip side (fixes #158)

### LINKED ISSUE #158: Letter animation should reveal the state of the guess only on the flip side

gheja:
See #158


gheja:
Fixed conflicts with current main.

PR #182, opened 2022-02-13
Implement Web Share API

igitur:
This PR implements the Web Share API when 
1. the device is on a mobile browser (implemented by a complicated regular expression)
2. and navigator.canShare() returns true
3. and navigator.share exists

I got the regular expressions from another open-source project on Github and I guess we'll just have to trust it.

The reason for the first requirement is that on Chrome for Desktop / Windows 10, the Web Share API is also available, but, on my PC at least, I can share it only to the default Windows 10 Mail app. I'd rather have it copied to the clipboard, like Wordle.


igitur:
I think this should be tested first on a number of devices before merged into `main`. 

Also, please consider rebasing rather than merging. The cleaner history makes it easier for downstream maintainers to update.

TheKnowles:
Tested on:

Chrome 98.0.4758.87
Android 10
Samsung Galaxy S9+

Looks good.

igitur:
So just note the idea is that the Web Share API is used only on mobile phones, not desktop browsers or *tablets*.  Results on the desktop and tablet will be copied to the clipboard.  I didn't make this choice explicitly. It seems to be the de facto choice to exclude tablets. For example, from http://detectmobilebrowsers.com/about :

> Android tablets, iPads, Kindle Fires and PlayBooks are not detected by design. To add support for tablets, add |android|ipad|playbook|silk to the first regex.

Should we support include tablets in the Web Share API ? I don't have a tablet to test or know what the typical user would expect.

gheja:
I have checked on Android 12, from Chrome 98.0 to Tweet, Messenger and a note application it works fine. From Firefox 97.1.0 it has problems, the Tweet is empty, Messenger does not show up and the note application shows "null" in the title.

I have also checked Wordle, the Share button does not bother to use the Share API on Firefox, it just copies to the clipboard.

The Share API is currently in Working Draft state so I think it is not surprising after all.

CanIUse reports around 59% support and 29% partial support: https://caniuse.com/web-share

src/components/modals/StatsModal.tsx
```diff
@@ -69,8 +69,12 @@ export const StatsModal = ({
...
             className="mt-2 w-full rounded-md border border-transparent shadow-sm px-4 py-2 bg-indigo-600 text-base font-medium text-white hover:bg-indigo-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500 sm:text-sm"
             onClick={() => {
-              shareStatus(guesses, isGameLost, isHardMode)
-              handleShare()
+              shareStatus(
+                guesses,
+                isGameLost,
+                isHardMode,
+                handleShareToClipboard
+              )
```
cwackerfuss, line 77: Nit: Is there a good reason to pass the callback into `shareStatus`? I would prefer this stay how it is since I think it's a little difficult to follow how/when the callback gets called. Maybe `handleShare` or `handleShareToClipboard` isn't a great name -- `showShareAlert` might be the better name for it.

src/App.tsx
```diff
@@ -280,7 +280,7 @@ function App() {
         gameStats={stats}
         isGameLost={isGameLost}
         isGameWon={isGameWon}
-        handleShare={() => {
+        handleShareToClipboard={() => {
           setSuccessAlert(GAME_COPIED_MESSAGE)
           return setTimeout(() => setSuccessAlert(''), ALERT_TIME_MS)
```
cwackerfuss, line 285: This just got merged tonight so I understand why it's not present in the PR, but can we switch to the new `showSuccessAlert` util function we're using? If you pull main, you should see uses of this throughout App.tsx.

cwackerfuss (changes requested):
Thanks for adding this, great idea to use the web share api!
Just a few comments.

src/App.tsx
igitur: I'll rebase on `main`.

src/components/modals/StatsModal.tsx
igitur: The share notification is shown only when sharing to the clipboard. When the Web Share API is used successfully, I didn't intend to show the notification. I think that's how Wordle also works and what would be most appropriate. Therefore I had to pass down the callback further, to be used where the distinction between the 2 methods is ultimately made.

igitur:
> I have checked on Android 12, from Chrome 98.0 to Tweet, Messenger and a note application it works fine. From Firefox 97.1.0 it has problems, the Tweet is empty, Messenger does not show up and the note application shows "null" in the title.
> 
> I have also checked Wordle, the Share button does not bother to use the Share API on Firefox, it just copies to the clipboard.
> 
> The Share API is currently in Working Draft state so I think it is not surprising after all.
> 
> CanIUse reports around 59% support and 29% partial support: https://caniuse.com/web-share

> I have checked on Android 12, from Chrome 98.0 to Tweet, Messenger and a note application it works fine. From Firefox 97.1.0 it has problems, the Tweet is empty, Messenger does not show up and the note application shows "null" in the title.
> 
> I have also checked Wordle, the Share button does not bother to use the Share API on Firefox, it just copies to the clipboard.
> 
> The Share API is currently in Working Draft state so I think it is not surprising after all.
> 
> CanIUse reports around 59% support and 29% partial support: https://caniuse.com/web-share

Really thorough testing, thanks! I'll try to reproduce and fix. If the API isn't available, I should be able to catch that and handle appropriately.

src/App.tsx
igitur: Rebased.

igitur:
> I have checked on Android 12, from Chrome 98.0 to Tweet, Messenger and a note application it works fine. From Firefox 97.1.0 it has problems, the Tweet is empty, Messenger does not show up and the note application shows "null" in the title.

Do you mean Firefox on *Android* specifically?

gheja:
> Do you mean Firefox on Android specifically?

Yes, I tested it on the same phone.

> If the API isn't available, I should be able to catch that and handle appropriately.

I'm afraid the API is available just not working as expected, as it is not raising an error but tries to share something that appears to be empty. But this is just a guess.

igitur:
I now just avoid the Web Share API when Firefox is in the user agent string and let results copy to the clipboard.

src/lib/browser.ts
```diff
@@ -0,0 +1,22 @@
+declare namespace window {
+  var opera: string
+}
+
+export const isMobileBrowser = function () {
+  const test = (ua: string): boolean => {
+    if (
+      /(android|bb\d+|meego).+mobile|avantgo|bada\/|blackberry|blazer|compal|elaine|fennec|hiptop|iemobile|ip(hone|od)|iris|kindle|lge |maemo|midp|mmp|mobile.+firefox|netfront|opera m(ob|in)i|palm( os)?|phone|p(ixi|re)\/|plucker|pocket|psp|series(4|6)0|symbian|treo|up\.(browser|link)|vodafone|wap|windows ce|xda|xiino/i.test(
```
JuanFontes, line 8: Hi @igitur did you copy this regex from `mobile-detect.js` or `Mobile-Detect`? I'm wondering about something, because according to their documentation:  
```
As mentioned later, "User-Agent" based detection is not a reliable solution in most cases, because:

The rules (regular expressions) are constantly outdated and incomplete
You have to update the detection code continuously
```

I'm not an expert, but what about using [screen.orientation](https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation)? Since smartphones support this and desktop browsers don't.

gheja:
I believe if the goal is to detect reliably then instead of hard coding a regular expression it would be better to use a popular and well maintaned package like [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) (11M downloads/week) that has a `getDevice()` method returning `{ model: '', type: '', vendor: '' }` where `type` is one of `console, mobile, tablet, smarttv, wearable, embedded`.

Note: I have also checked [react-device-detect](https://www.npmjs.com/package/react-device-detect) (480k downloads/week) and it is using this package. [mobile-detect](https://www.npmjs.com/package/mobile-detect) (200k dl/w) on the other hand does not depend on it, but was updated about a year ago.

src/lib/browser.ts
JuanFontes: > I believe if the goal is to detect reliably then instead of hard coding a regular expression it would be better to use a popular and well maintaned package like [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) (11M downloads/week) that has a `getDevice()` method returning `{ model: '', type: '', vendor: '' }` where `type` is one of `console, mobile, tablet, smarttv, wearable, embedded`.
> 
> Note: I have also checked [react-device-detect](https://www.npmjs.com/package/react-device-detect) (480k downloads/week) and it is using this package. [mobile-detect](https://www.npmjs.com/package/mobile-detect) (200k dl/w) on the other hand does not depend on it, but was updated about a year ago.

Hi @gheja I've checked this package too, and at the end, they use a regex https://github.com/faisalman/ua-parser-js/blob/master/src/ua-parser.js#L206 -- At least we would not need to update the code, just the package version if something breaks.

gheja: Yes, that's what I meant. That project has a lot of contributors and as it is being used by a lot of other projects, I'd assume that they have much wider range of devices to test and they would react much quicker to a change than we ever could here.

igitur:
> a `getDevice()` method returning `{ model: '', type: '', vendor: '' }` where `type` is one of `console, mobile, tablet, smarttv, wearable, embedded`.

Let me try that. But let's decide. For which of those device type values are we going to attempt to use the Web Share API. Only `mobile`?

And we'll still have to exclude Firefox.

pietroppeter:
not sure if it is helpful since the goal of this PR seems to be different, but I tested again with the latest deployed version (erj5n...) the issue #206 and this PR does not fix it. In that same issue I share the code of a wordle instance derived from this repo many commits ago where the implementer had a different way of sharing that works when tested with the same iPhone6/iOS as reported in #206. For convenience I report here the details:

> There is another italian version (parolette) which is based on an earlier version of this clone: https://github.com/xho/parolette
That version was forked before share was implemented and so share was implemented in a different way. I can confirm that with the above mentioned phone/os/browser sharing with parolette works, while (as mentioned) it does not work with this clone.

> It seems to me the difference resides in [this line](https://github.com/xho/parolette/blob/d22f3d6541056abf6b9caf38eeaa54491700a75b/src/components/elements/ShareButton.tsx#L41) which offers an alternative share mechanism compared with [what is currently implemented](https://github.com/cwackerfuss/react-wordle/blob/5660d14f199675ae147eb601f21c8ec733a76ef3/src/lib/share.ts#L11) here.

igitur:
https://github.com/xho/parolette uses react-device-detect, which depends on ua-parser-js, which has been mentioned above as a possible solution.

src/lib/browser.ts
igitur: > I'm not an expert, but what about using [screen.orientation](https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation)? Since smartphones support this and desktop browsers don't.

`screen.orientation` is supported on the desktop. https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation#browser_compatibility

Try it out in the console

gheja:
> For which of those device type values are we going to attempt to use the Web Share API.

I am not sure, honestly. Maybe keeping a list with known working configurations would be the safest bet for the moment (I believe the support is growing) and the clipboard fallback is still a good way to share, in my opinion.

If you asked me I'd go with the safe version, as we can always come back and make a PR with an improved one.

igitur:
I created a separate PR #233 to test the implementation of the ua-parser-js library to detect the user's device details. Please test that one too. Hopefully #233 gets merged, and then we can abandon this PR.

pietroppeter:
yep, will test and report back! thanks!

PR #191, opened 2022-02-14
Refactored alerts

cwackerfuss:
With the number of alerts we had, I noticed there was an opportunity to refactor these out. Instead of using `useState` in App.tsx, we can use a React context provider and interact with the new `useAlert` hook to declaratively render alerts without having to worry about as much about the mount/unmount process.


src/App.tsx
```diff
@@ -140,23 +136,22 @@ function App() {
...
+      const delayMs = REVEAL_TIME_MS * MAX_WORD_LENGTH
 
-        setTimeout(() => {
-          setSuccessAlert('')
-          setIsStatsModalOpen(true)
-        }, ALERT_TIME_MS)
-      }, REVEAL_TIME_MS * MAX_WORD_LENGTH)
+      showSuccessAlert(winMessage, {
+        delayMs,
+        onClose: () => setIsStatsModalOpen(true),
```
cwackerfuss, line 145: I added the `onClose` option to account for this case. The alert is closed automatically after `ALERT_TIME_MS` (by default)

TheKnowles:
Nice job! This is an elegant solution.

abrarhayat (approved):
The Alert Context implemented here is a great display of the usage of react hooks!

TheKnowles (commented):
Wow I did not realize this was not submitted. Sorry for that.

PR #201, opened 2022-02-14
Duplicate letters check fix

gheja:


src/lib/words.ts
```diff
@@ -16,27 +16,45 @@ export const isWinningWord = (word: string) => {
...
+      knownLetterArray.push(guess[i])
+    }
+    if (statuses[i] === 'correct' && word[i] !== guess[i]) {
         return WRONG_SPOT_MESSAGE(guess[i], i + 1)
-      }
     }
   }
 
-  for (const letter of Array.from(knownLetterSet.values())) {
+  for (const letter of knownLetterArray) {
```
TheKnowles, line 38: Because the newly added duplicate check also works for stepping through the letters to confirm usage, we could drop this loop entirely.

TheKnowles:
I like the letter strip solution. One note, otherwise looks good.

src/lib/words.ts
gheja: Oh, you're right, I will update the PR.

cwackerfuss (approved):
Thanks @TheKnowles for reviewing. Tested locally and works good.

Issue #203, opened 2022-02-15
MAX_WORD_LENGTH - Issue with length of string in Unicode characters


muhelen:
When using unicode character like "மு" (\u0BAE\u0BC1), the length return as 2, but the "actual"length should be 1.  This affected the MAX_WORD_LENGTH throughout the code.

How to fix this?

gheja:
Unfortunately a "letter" is not always one unicode character and it is hard to parse them properly.

The best to do this would be using a well maintaned library like [grapheme-splitter](https://github.com/orling/grapheme-splitter), copied directly from that readme:
```javascript
var splitter = new GraphemeSplitter();

// plain latin alphabet - nothing spectacular
splitter.splitGraphemes("abcd"); // returns ["a", "b", "c", "d"]

// two-char emojis and six-char combined emoji
splitter.splitGraphemes("🌷🎁💩😜👍🏳️‍🌈"); // returns ["🌷","🎁","💩","😜","👍","🏳️‍🌈"]

// diacritics as combining marks, 10 JavaScript chars
splitter.splitGraphemes("Ĺo͂řȩm̅"); // returns ["Ĺ","o͂","ř","ȩ","m̅"]

// individual Korean characters (Jamo), 4 JavaScript chars
splitter.splitGraphemes("뎌쉐"); // returns ["뎌","쉐"]

// Hindi text with combining marks, 8 JavaScript chars
splitter.splitGraphemes("अनुच्छेद"); // returns ["अ","नु","च्","छे","द"]

// demonic multiple combining marks, 75 JavaScript chars
splitter.splitGraphemes("Z͑ͫ̓ͪ̂ͫ̽͏̴̙̤̞͉͚̯̞̠͍A̴̵̜̰͔ͫ͗͢L̠ͨͧͩ͘G̴̻͈͍͔̹̑͗̎̅͛́Ǫ̵̹̻̝̳͂̌̌͘!͖̬̰̙̗̿̋ͥͥ̂ͣ̐́́͜͞"); // returns ["Z͑ͫ̓ͪ̂ͫ̽͏̴̙̤̞͉͚̯̞̠͍","A̴̵̜̰͔ͫ͗͢","L̠ͨͧͩ͘","G̴̻͈͍͔̹̑͗̎̅͛́","Ǫ̵̹̻̝̳͂̌̌͘","!͖̬̰̙̗̿̋ͥͥ̂ͣ̐́́͜͞"]
```

[...]

PR #205, opened 2022-02-15
Unicode character handling fix

gheja:
Fixes the issue when using letters using more than one unicode characters - see issue #203 

I updated the length check, display in grid, enter and delete operations as well. It looks good to me but probably needs more testing.

I have modified the keyboard to have unicode characters, like this:
```patch
diff --git a/src/components/keyboard/Keyboard.tsx b/src/components/keyboard/Keyboard.tsx
index 485e8b3..5788dd4 100644
--- a/src/components/keyboard/Keyboard.tsx
+++ b/src/components/keyboard/Keyboard.tsx
@@ -77,7 +77,7 @@ export const Keyboard = ({
         <Key width={65.4} value="ENTER" onClick={onClick}>
           {ENTER_TEXT}
         </Key>
-        {['Z', 'X', 'C', 'V', 'B', 'N', 'M'].map((key) => (
+        {['Z', 'X', 'C', 'V', 'B', 'N', '\u0BAE\u0BC1'].map((key) => (
           <Key
             value={key}
             key={key}
```


cwackerfuss (approved):
This looks awesome and works well after testing locally. Thanks for this @gheja

cwackerfuss:
Oops - it looks like there is a merge conflict with main. Can you please address this and then we can merge

muhelen:
Its still didn't fix the MAX_WORD_LENGTH issue. The unicode characters still count as separate characters, which in return affecting the length number.

muhelen:
Now **NOT_ENOUGH_LETTERS_MESSAGE** working fine. The unicode characters combination could be up to 5 letters, but the last letter couldn't combine fully. For example, it could enter as **ம** but ம+் which in return supposed to be ம் not working..

muhelen:
The last grid don’t combine Tamil unicode  characters. Anyone else having this issue on other languages?

gheja:
Good to hear, I am glad it works.

Thank you so much, and all the best for you too!

@cwackerfuss, I think we have fixed everything, could you check it? Thanks!

cwackerfuss:
Looks awesome thanks all!

PR #207, opened 2022-02-15
add plausible analytics

pietroppeter:
this adds the possibility to use plausible analytics instead of google analytics. not sure if it can be helpful for others but I am using this here: https://github.com/par-le/react-wordle


cwackerfuss (approved):
Great idea, thanks!

Issue #188, opened 2022-02-13
Hard to use on mobiles with small screens


oradwell:
**Describe the bug**
Need to scroll on iPhone's with small screens

**To Reproduce**
Open the page on iPhone SE (2020)

**Expected behavior**
The whole screen should be utilised, and there should be no empty space

**Screenshots**
![Screenshot from 2022-02-13 21-06-45](https://user-images.githubusercontent.com/5054896/153775116-777654ba-44dc-4c6b-9798-810303fdd64f.png)
![Screenshot from 2022-02-13 21-11-05](https://user-images.githubusercontent.com/5054896/153775210-eda5f4f7-faeb-4c05-ae0f-114012e13619.png)

**Smartphone:**
 - Device: iPhone SE (2020)
 - OS: v13.4
 - Browser: Safari

**Additional context**
We should resize the elements on the page to fit them on all screen sizes.

candardo:
I get the same behaviour on Android 12 with "large text" accessibility setting on:

- in webview
- on Firefox either using system accessibility setting or its own zoom text setting

No issue on Chrome.

oradwell:
this is fixed by #209

PR #209, opened 2022-02-16
improve layout of header

cwackerfuss:
I've been unhappy with how much margin there is currently at the top of the page. This reduces the margin and improves the positioning of the alert.

I also removed the SVG hover animation since it felt a little awkward and occasionally bugs out the cursor when you're hovering on the very bottom of the icon.

Screen Shot 2022-02-15 at 10 58 58 PM

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -246,7 +246,7 @@ function App() {
   }
 
   return (
-    <div className="pt-2 pb-8 max-w-7xl mx-auto sm:px-6 lg:px-8">
+    <div className="pt-2 pb-8 max-w-7xl mx-auto sm:px-6 lg:px-8 h-screen flex flex-col">
       <div className="flex w-80 mx-auto items-center mt-4 mb-6">
         <h1 className="text-xl ml-2.5 grow font-bold dark:text-white">
           {GAME_TITLE}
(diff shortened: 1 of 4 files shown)


diondiondion:
This is much better!

However, I have a very tall phone and the keyboard is quite high up now. Could you add some CSS that makes sure the keyboard is pushed to the bottom of the page when there's space? If the outer container matches the height of the viewport, making it a flex (column) container and adding `margin: auto 0` to the grid in the middle may do the trick.

gheja:
There is one more open issue with playing on smaller screens: #188

I've been thinking about adding CSS Media Queries to modify the margins based on screen size (or aspect ratio?) but haven't got the time to create a set with reasonable cover of the possible cases.

zbennett:
What is the latest with this PR? I really like this change.

cwackerfuss:
Hi @zbennett - I can take a look at this when I'm back tomorrow - had a vacation this week. :)

zbennett:
I love every part of that statement.

cwackerfuss:
Added a flex column layout to the outermost app container of 100vh and made the guesses section flex-grow to fill space - just merged thanks all!

vegeta897:
Unfortunately 100vh is problematic on iOS with Chrome and Safari.

image image

If there was a good solution out there or if I was more familiar with react and tailwind I'd try to make a PR, but for now I just wanted to bring this to your attention. I can open an issue if that would be helpful.

PR #211, opened 2022-02-16
Improved the animations. Hopefully this won't glitch

abrarhayat:
The previous animation we had for the svg was a little bit glitchy and hence I am suggesting some improved animations. As a bonus I have added some animations to the keyboard.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -164,18 +164,6 @@
   text-shadow: 1px 1px 1px #000000;
 }
 
-@keyframes float {
-  0% {
-    transform: translate(0, 0);
-  }
-  50% {
-    transform: translate(0, -0.5rem);
-  }
-  100% {
-    transform: translate(0, 0);
-  }
-}
-
 @keyframes jiggle {
   0% {
     transform: translate(0, 0);
--- a/src/index.css
+++ b/src/index.css
@@ -53,6 +53,14 @@
   animation-fill-mode: backwards;
 }
 
+svg.cursor-pointer {
+  transition: all var(--animation-speed-fast);
+}
+
+svg.cursor-pointer:hover {
+  transform: scale(1.2);
+}
+
 .jiggle {
   animation: jiggle linear;
   animation-duration: var(--animation-speed-fast);


diondiondion:
The keyboard animation causes horizontal scrollbars when tapping a letter at the right edge of the keyboard. In general I think the game can do without these animations, they don't really add much.

abrarhayat:
Removed all keyboard animations. Just trying things out, Let's hear from what others have to say.

cwackerfuss:
I don't mind the scale animation when hovering these SVG icons, but 1.5 is a little much. I like subtle animations - could we try 1.2 or 1.1?

And I'm a _little_ worried about fork maintainers running into issues if we apply this animation to all SVGs. Can we give these svg icons a specific class that applies the animation instead?

abrarhayat:
Reduced the scaling animation to 1.2 and now pointing animations to svgs with cursor-pointer class.

gheja:
IMHO this looks much cleaner and better than the bouncing animation. Also, the bouncing was a bit odd when the button kept the :hover and it was jumping continously in the background while a modal was open (on Andorid + Chrome).

ericdrgn:
I actually dislike these animations after testing them extensively compared to the old bounce. FWIW if you change the old bounce animation to linear from infinite it won't keep doing the animation. 

I just think that this growing animation is is a little much and isn't as clear an animation as the old one. Honestly I don't care if this gets merged into main, I just won't use it on my fork(s), love seeing improvements keep rolling just don't think this one looks as good as the bounce. Since transitioning to linear it doesn't glitch either

abrarhayat:
After you suggested, I did try out the float linear animation and in my opinion, the float animation only looked good when it was running infinitely. However, several times, while I was testing out the buttons, I had faced glitches like @cwackerfuss already mentioned in the header layout improvement PR. Hence I am suggesting to replace it with something much cleaner and subtle which is also not a boring old static icon.

abrarhayat:
Added a pulsing animation. Will remove this commit if its buggy. There will remain the issue in mobile view, where the animation will keep going on behind the modals.

ericdrgn:
@abrarhayat I do like the pulse animation a lot better, but still think infinite is excessive especially since it keeps going behind modals forever. Have you played with just doing something like ```animation: pulse 3;``` that way it just runs for as many loops as you want it to. So even on mobile it will still run behind a modal but eventually stops.

abrarhayat:
In that case, I'd rather go back to the basic scale animation. The entire point of this pulse animation was so that it would go on. But then again, there'll be a bug on mobile. So I think we'd be better off with just scale.

cwackerfuss:
Agreed with @diondiondion -- can we nix the pulse and just have a subtle scale animation please?

abrarhayat:
Removed the last commit. Now we just have the subtle 1.2x scale animation.

abrarhayat:
Here is the preview for the current state:

https://react-wordle-jm4ftcajs-cwackerfuss.vercel.app/

cwackerfuss:
Looks great - thanks @abrarhayat !

PR #221, opened 2022-02-17
Refactor emoji grid dependency parameters

igitur:
The resolution of `isDarkMode` and `isHighContrast` to its values shouldn't be deep inside the method that builds the share text. Rather the variables that are already in use higher up in the stack should be passed down.

Also factored out a new method to return the emoji tiles to be used to build the final text.


gheja:
I like the idea of basically using a "palette" and not doing logic everywhere.

cwackerfuss (approved):

PR #229, opened 2022-02-18
[POC] - Make MAX_WORD_LENGTH  dynamic based on getWordOfDay - DO NOT MERGE

JuanFontes:
**Disclaimer**: I'm not a developer, I'm just an enthusiast and this is a POC, feel free to correct any error I've made.

I was thinking about creating a new project based on something (I put Disney characters as an example) but I have different words with different lengths. So I tried to make it dynamic, I did some tests and it worked every time.

I'm wondering if this is a good idea that we can have on `react-wordle`, if so, I can work more to improve and make it better. I appreciate any feedback.

Notes:
* [src/constants/wordlist.ts](https://github.com/cwackerfuss/react-wordle/pull/229/files#diff-cc11204929d6665f6025e21218c55c09af08ba0892ec2f28c1f62efc1d7157a2) - **mocked for testing purposes, feel free to change the words if you want to test it too. It will be reverted later.**


JuanFontes:
hi @ericdrgn thanks for the feedback! :) - I sent a new commit fixing this typo, let me know if it worked

ericdrgn:
@JuanFontes yup all fixed now!

aghasemi:
Hi. I created a PR with the same intent: https://github.com/cwackerfuss/react-wordle/pull/252. It is already implemented (and working, apparently) in two variations. Here is the PR: https://github.com/cwackerfuss/react-wordle/pull/252

ericdrgn:
@JuanFontes I just tested out @aghasemi's implementation as well and that one from PR #252 is much better. While both accomplish the same thing, the method they implemented requires far less changes in the code while getting the same result. Your work is still appreciated and thi simpler solution was one I hadn't thought of either!

JuanFontes:
Hi @ericdrgn! It was a pleasure reading your comment, thanks for all the words! As I said before, I'm not a developer, it was my first time working with js/react, I'm just an enthusiast. I know the solution is really "ugly" even though it's working hehe. It was just a POC that could inspire other solutions.

The approach on #252 is really better, I have just some concerns like: Fixing a "constant" value/file as dynamic, maybe there is a better way to do that, but I'll leave it for the specialists. 
 
Thank you again Eric!

Issue #206, opened 2022-02-15
Share button does not work (iPhone6, Safari, iOS 12.5.5) + hints for solution


pietroppeter:
**Describe the bug**

Share button does not copy results. I know this has been reported multiple times and ascribed to HTTP (#94 and #43) but I think I have new elements and a possible hint for a solution, see below in context ("To Reproduce" and "Expected behavior" can be skipped).

**To Reproduce**

1. solve the puzzle
2. click share
3. no modal for share successful appears
4. pasting in a chat or elsewhere does not paste the correct result

The behavior was found with an iPhone 6 (iOS 12.5.5) using Safari 

**Expected behavior**

3. modal confirm copying to clipboard should appear
4. pasting should paste the correct result

**Smartphone (please complete the following information):**
 - Device: iPhone6
 - OS: 12.5.5
 - Browser safari
 - Version ?

phone was not mine but I was able to confirm the issue

**Additional context**

I am starting to port my Italian version (born on Jan 3rd, same day when this clone was born) from the usage of original world files (which are not licensed open source) to use this clone. My current italian version is here (not relevant for this): https://github.com/pietroppeter/wordle-it 

I have set up a new version (translation ongoing): https://github.com/par-le/react-wordle

There is another italian version (parolette) which is based on an earlier version of this clone: https://github.com/xho/parolette
That version was forked before share was implemented and so share was implemented in a different way. **I can confirm that with the above mentioned phone/os/browser sharing with parolette works, while (as mentioned) it does not work with this clone**.

It seems to me the difference resides in [this line](https://github.com/xho/parolette/blob/d22f3d6541056abf6b9caf38eeaa54491700a75b/src/components/elements/ShareButton.tsx#L41) which offers an alternative share mechanism compared with [what is currently implemented](https://github.com/cwackerfuss/react-wordle/blob/5660d14f199675ae147eb601f21c8ec733a76ef3/src/lib/share.ts#L11) here.

I will try to make the share approach of parolette work with the current version of this clone, but since I have never coded in typescript that might take me a moment (if anyone wants to give pointers happy to take them :)).

Thanks for the great work on this and have a good day.

cwackerfuss:
Hey @pietroppeter thanks for reporting. Can you try the share functionality from this link?

https://react-wordle-git-fork-igitur-implement-share-api-cwackerfuss.vercel.app/

[...]

PR #233, opened 2022-02-18
Implement Web Share API (and use ua-parser-js)

igitur:
Alternative PR to #182 that uses ua-parser-js to determine whether the user is using a mobile device instead of maintaining our own complex regex.


cwackerfuss:
This looks good - @igitur can you fix the conflicts on this and then I can approve?

igitur:
I already included this in my Wortel implementation. Got reports last night that sharing directly to Facebook doesn't work. Sharing to eg WhatsApp does work as intended. So I just need to look into that a bit. Anyone that can confirm the Facebook issue?

Will rebase.

igitur:
> Tested to see if this fixes #206 but currently it does not

I think we'll have to park that and address later. Let's get the rest of this PR in first.

> Got reports last night that sharing directly to Facebook doesn't work. 

So when the share snippet contains a valid URL (including `https://`) this totally confuses Facebook and Facebook tries to post only the URL and ignores the rest of the snippet. In my implementation, I just removed the `https://`, e.g. 

```
🥕 Wortel 20 2/6

🟧⬜🟩⬜⬜
🟩🟩🟩🟩🟩

wortel.wrintiewaar.co.za
```

gheja:
> when the share snippet contains a valid URL (including https://) this totally confuses Facebook

So then that's the reason why Wordle has not included a URL in the first place! I was wondering but this sounds reasonable. And also, Twitter/Facebook would add a card as well, stealing the focus from the results (and spamming the timeline (even more?)).

igitur:
@cwackerfuss Rebased and happy from my side that this PR is complete.

gheja:
In issue #261 it is pointed out that Facebook and Messenger built-in browsers have limitations, i.e. no proper cookie/local storage handling and neither webshare nor copy to clipboard work.

I think it might be a good idea to use ua-parser-js to warn the user about this, as the game will be limited and their progress will be lost when they play this way.

sourencho:
Why was this PR closed?
It would be awesome to have Web Share API!

igitur:
Good question! @cwackerfuss , what happened here? Was it closed accidentally?

igitur:
Sorry, I deleted my fork and didn't realise that it will close this PR. I'll resubmit this.

igitur:
Resubmitted at #289.

Thanks for picking this up @sourencho.

Issue #218, opened 2022-02-16
Harder mode (feature suggestion)


gheja:
An even harder mode where the "present" clue is enforced more strictly, the guess must not contain the letter in the position that led to a "present but not here" before.

pietroppeter:
relevant tweet by Josh Wardle on this harder mode: https://twitter.com/powerlanguish/status/1479226926636089347?s=20&t=SlTr_5kspXEQCTA53M_cag

> knowing what I know now, I might go with your interpretation instead

gheja:
Thanks! Yes, adding a "you can't use the \"absent\" letters" rule would be great, too.

And I must admit I forgot that I played earlier by this custom rule as well - to increase the efficiency, but still I finished with 4 or 5 guesses mostly.

cwackerfuss:
@gheja for me, this feels like one of those features that, since it isn't in the OG wordle, probably best to not include it in the core project here. Thoughts?

I wish there was a clean way to include these "optional" features in the build. Maybe we need to build a scaffolding/generator cli tool... :D

gheja:
I'm on the fence, I'd suggest this as an optional feature so it would not affect the game play by default.

But again, every new (even optional) feature would make testing more and more complicated, the codebase more prone to accidental errors.

I'd totally agree if you discarded this, defining the scope and focusing on that is important, and this is definitely not part of the OG Wordle. On these lines, I'd discard it too.

p.s. I really need to start playing with the automated tests :)

PR #236, opened 2022-02-18
Add timer

alexaylwin:
This PR adds a simple timer to the game. It has the following changes:
- Adds a `time` state variable to the `App` component
- Adds  a `hasGuessingStarted` state variable to the `App` component
- Adds a `Timer` component which updates the `time` once per second and displays it
- Adds an "Average Time" stat
- Adds the time for the current game in the 'Share' button


gheja:
Sounds great, I wonder why this hasn't been implemented earlier.

Just a few things I noticed:
  * I think the timer is too prominent on the GUI
  * it only counts in seconds, so 2 minutes is shown as 120s
  * the timer is paused while entering the word (spam the keyboard to "pause" while you think)
  * when reloading the page the timer gets reset to 0 and stucks there (but the guesses persist and you can continue the game)

Also, not a technical note but: now it is clearly visible how much time I spend playing the game... :D

On this note, addig a sum of time played to the stats modal would be fun?

Tafiti:
This is a great feature. Along the same lines, I wonder if this can be extended to include a countdown mode. This creates a "timed-mode", if you will, which can be both custom or defaults. A 10-minutes default, for example, starts when when one starts the first guess and ends either when the word is solved for, or when the timer gets to 0:00:00. 

I think this PR comes closest to implementing this feature.

gheja:
Although I think the play time depends on the game style as well. In hard mode (or [harder mode](https://github.com/cwackerfuss/react-wordle/issues/218) ) it can take much more time, some days I even started playing Wordle in the morning, got back to my things and finished it at night.

If the timer is enforced it should be optional, in my opinion.

Tafiti:
@gheja yes, it should be optional, just like the other modes are. I play this in a family group and have been thinking how we can make it more interesting. It is already interesting as it is, but just thinking how we can get everyone to play with intent.

alexaylwin:
Thanks for the comments. If people actually think this feature would be fun, I'll pull this PR back and make the following changes:

- Hide the timer behaviour behind a "mode" toggle
- Fix defects related to the timer pausing
- Clean up the time presentation (add mm:ss, dark mode, font)
 
- In terms of when the timer starts/stops - I'd like it to be continuous from the first submitted guess. If you walk away for 9h and come back, the clock keeps ticking. My rationale is that during that time you may still be figuring out words to play.

It also makes for a slightly cleaner/more performant implementation because I can just calc the delta from a time started timestamp, vs. write to local storage every 1000ms

ericdrgn:
Would there be a way to toggle it in coordination with hard mode? So if you are in hard mode the timer runs continuously from the first guess but if you are in standard/easy mode it only runs if/when you are actively in the app? Of course all this taking into account it being optional as well.

So with timer toggle off nothing happens, and timer can only be turned on before first guess like hard mode, but then when it is on it depends on what mode you are in.

- Hard mode & timer on: timer doesn't stop once you enter your first guess
- Easy mode & timer on: timer continues only when actively in the game

ericdrgn:
An alternative is to just make it not hidden in the menu and it function similarly to a chess timer. Where it is entirely up to the user how it is used and turning it on starts the timer and turning it off pauses it. Then it just resets to zero on a new day. So if as @Tafiti mentioned a family group (or just any group) decides that they will play with it running continously they can. Or if they pause it when they walk away than they can and that is just how a group decides it should go. 

This way leaves it entirely to the user and more on the honor system of course but I don't think that is necessarily a bad thing.

Anyway those are the two thoughts I have on it, I like the idea either way and think however it works out it is a cool addition.

gheja:
Just a quick comment:

> I'll pull this PR back

You don't have to. You can keep working on your original branch and the commits will show up here in the PR. (I think it would be better and cleaner if you kept working on this PR, but this is just my preference.)

src/App.tsx
```diff
@@ -256,6 +269,7 @@ function App() {
           onClick={() => setIsSettingsModalOpen(true)}
         />
       </div>
+      <Timer time={time} started={hasGuessingStarted} onTick={onTick}></Timer>
```
cwackerfuss, line 272: Let's remove the `hasGuessingStarted` concept and just pass `guesses.length > 0` here -- this value is always derived directly from `guesses` and using `useState` feels wrong.
If you don't like the inline logic I suggested, you could also do:
```
const hasGuessingStarted = guesses.length > 0
```
somewhere at the top of the component.

src/App.tsx
```diff
@@ -237,6 +246,10 @@ function App() {
     }
   }
 
+  const onTick = () => {
+    setTime((prevTime) => prevTime + 1)
+  }
```
cwackerfuss, line 251: This doesn't persist when you refresh the page. You could update `StoredGameState` localstorage value to include elapsed time: https://github.com/cwackerfuss/react-wordle/blob/main/src/lib/localStorage.ts#L4

cwackerfuss (changes requested):

src/App.tsx
cwackerfuss: I just noticed you mentioned an implementation where you store "time started" instead of the elapsed time in localstorage. I think that's smart!

cwackerfuss:
@alexaylwin I agree with @gheja on his points too, btw -- right now it takes up a little too much space on the top of the page. Since this is meant to remain a faithful clone of the original game, I also don't want to introduce any new UI elements to the main view that a fork maintainer is not expecting -- there are quite a few live projects that continue to pull in updates we make.

I think the best place to put the time it took to complete the puzzle is either in the Win alert or the Stats modal.

alexaylwin:
@cwackerfuss sure thing. I'll make some more commits addressing your and others' comments. If you're concerned about this drifting too far from the original game, I'm good to just leave it as a fork. Just thought it'd be a neat addition, and others would might find it interesting to build on.

alexaylwin:
Or, I could make a new PR with only the "infrastructure" changes to measure and track time + time stats, and remove the presentation entirely? Leave it up to others to choose if they want to access and use the data.

gheja:
I think it would be great in the Stats page so you can see it when you finished. Or maybe an option that can be turned on - that way it is up to the users if they want to see it on the game screen. This might turn the feature on/off entirerily, I am sure there would be "speed runners" who compare their Stats screens on a daily basis, I have seen some sharing it on Twitter. (This of course would complicate the thing of how many games were played vs. how many were played with the timer on.)

cwackerfuss:
Going to close this for now @alexaylwin for housekeeping- if you end up refactoring please feel free to open another PR - thank you!

PR #241, opened 2022-02-19
Secure production docker container: Run nginx with non root user

naeem-akhtar:
Running a docker container in production with a root user is not a good idea, since the root user can access a broader range of kernel services.

Following are some examples from [Stackoverflow](https://stackoverflow.com/questions/19054029/security-of-docker-as-it-runs-as-root-user).
- Manipulate network Interfaces, Routing tables, Netfilter rules.
- Create raw sockets (and generally speaking, "exotic" sockets, exercising code that has received less scrutiny than good old TCP and UDP).
- Mount/unmount/remount filesystems.
- Change file ownership, permissions, extended attributes, overriding regular permissions (i.e. using slightly different code paths);
etc.

### Changes in PR:
- Modified the Nginx configs to allow non-root users to run Nginx.
- Added custom 'reactle' user that will run the container.
- Exposed port 8080 of docker container since the non-root user cannot bind port below 1024.
- Updated the command to build and run docker containers.

### DIFF
--- a/docker/build_system.sh
+++ b/docker/build_system.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+# Strict shell settings.
+set -e pipefall;
+set -u pipefall;
+
+UID=1001
+GID=1001
+USER=reactle
+GROUP=reactle
+
+# Enable print for all executed commands
+trace_on() { set -x;}
+# Disable print for all executed commands
+trace_off() { set +x;}
+
+add_custom_user_group() {
+    addgroup -g $GID -S $GROUP
+    adduser -S -D -H \
+        -u $UID \
+        -G $GROUP \
+        -s /sbin/nologin \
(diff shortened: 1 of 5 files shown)


docker/build_system.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
+# Strict shell settings.
+set -e pipefall;
+set -u pipefall;
+
+UID=1001
+GID=1001
+USER=reactle
```
gheja, line 9: Please note that USER is an environment variable already set, but I believe this won't cause any problems in this short script.

naeem-akhtar: @gheja Thanks for pointing out. I will update the variable name in future PR. For now this will not affect the script but it might when more code will be added.

cwackerfuss (approved):

Issue #239, opened 2022-02-19
Character problem in words like "ç" and "i"


Cryptoseen:
I'm just learning coding for hobby purposes. I get an error when I replace all of the wordlist.ts with Turkish words containing Turkish letters such as ç,i and build it. If I change only today's word in wordlist.ts, it builds and there is no problem. But this time, if it is a Turkish word with the letter i or ç in it, I need to use it like I or C in English. When I try with ç and i it gives the error word not found. Where am I doing wrong?

**Additional context**

gheja:
The code [checks for exact matches](https://github.com/cwackerfuss/react-wordle/blob/main/src/lib/words.ts#L6-L15) so you need to update [the keyboard](https://github.com/cwackerfuss/react-wordle/blob/main/src/components/keyboard/Keyboard.tsx) to have those specific keys.

Cryptoseen:
> The code [checks for exact matches](https://github.com/cwackerfuss/react-wordle/blob/main/src/lib/words.ts#L6-L15) so you need to update [the keyboard](https://github.com/cwackerfuss/react-wordle/blob/main/src/components/keyboard/Keyboard.tsx) to have those specific keys.

https://imgur.com/bU9UTN8   This is how I added the special characters in the beginning. When I add words containing Turkish language to the Wordlist file and compile, I get the following error:

    'Words' are not exported from '../constants/wordlist' (imported as 'Words').


But the wordlist file remains original and when I just change the word of the day the compilation doesn't give an error. But at that time I am getting "word not found error" for word containing letters ç and i. There is no problem with Turkish characters like "ş".

gheja:
I have added a few dummy words with the letters you provided ('aaaaç' 'aaaai' 'aaaaş'), updated the keyboard and it works for me.

Could you check the patch below?

![image](https://user-images.githubusercontent.com/1894642/154814036-8f257246-858e-4320-a733-21161bf6dca9.png)

```patch
diff --git a/src/components/keyboard/Keyboard.tsx b/src/components/keyboard/Keyboard.tsx
index 485e8b3..bb88e9b 100644
--- a/src/components/keyboard/Keyboard.tsx
+++ b/src/components/keyboard/Keyboard.tsx
@@ -77,7 +77,7 @@ export const Keyboard = ({
         <Key width={65.4} value="ENTER" onClick={onClick}>
           {ENTER_TEXT}
         </Key>
[...]

PR #242, opened 2022-02-19
Locale aware uppercase lowercase fix

gheja:
The original version uses `toLowerCase()` and `toUpperCase()` that are not locale aware, see Issue #239

For example the letter "i" is different in Turkish.
```
> 'i'.toUpperCase()
< 'I'

> 'i'.toLocaleUpperCase('tr-TR')
< 'İ'

> 'i'.toUpperCase().charCodeAt(0)
< 73

> 'i'.toLocaleUpperCase('tr-TR').charCodeAt(0)
< 304
```

I added a setting `LOCALE_STRING` that (if set) makes the code use `toLocaleLowerCase()` and `toLocaleUpperCase()` with the correct locale.

Thanks for the help and testing @Cryptoseen and @joedel263!


src/constants/settings.ts
```diff
@@ -3,3 +3,4 @@ export const MAX_CHALLENGES = 6
 export const ALERT_TIME_MS = 2000
 export const REVEAL_TIME_MS = 350
 export const GAME_LOST_INFO_DELAY = (MAX_WORD_LENGTH + 1) * REVEAL_TIME_MS
+export const LOCALE_STRING = ''
```
cwackerfuss, line 6: Thoughts on putting this in `.env` as `REACT_APP_LOCALE_STRING`?

Think it makes more sense than in settings.ts

cwackerfuss (changes requested):
Looks good! Just one piece of feedback -- I think it makes more sense in .env

src/constants/settings.ts
gheja: Sounds reasonable, I will move it there.

gheja: Moved it to .env.

cwackerfuss (approved):

PR #252, opened 2022-02-21
Update settings.ts to allow solutions of varying sizes

aghasemi:
Set `MAX_WORD_LENGTH` and `MAX_CHALLENGES` dynamically based on the solution of the day, allowing words of varying sizes on different days.

This change allows having a word list composed of words of varying sizes, without having to change the setting each day for a word of a new size. This is particularly useful for languages/games where the number of possible solutions with a fixed size is too small to allow for a reasonable game to be designed.


aghasemi:
Any idea how I can see the Vercel bot's failure logs/reason?

4thehalibit:
> Any idea how I can see the Vercel bot's failure logs/reason?

Are you looking for something like this https://vercel.com/guides/debugging-and-troubleshooting-vercel-logs-with-sematext

cwackerfuss:
Hey @aghasemi - I'm not in love with this approach. It seems overengineered for a simple variable setting. Going to close this for now. Thank you though!

travisryan:
This worked great!

cwackerfuss:
@aghasemi I've been swayed - seems like people are liking this approach and the more I think about it, the more it makes sense. Sorry about closing!
I'd like to reopen this and if there are no conflicts, approve and merge.

src/constants/settings.ts
```diff
@@ -1,6 +1,7 @@
-export const MAX_WORD_LENGTH = 5
-export const MAX_CHALLENGES = 6
+import { solution } from '../lib/words'
+
+export const MAX_WORD_LENGTH = solution.length
+export const MAX_CHALLENGES = MAX_WORD_LENGTH + 3
```
cwackerfuss, line 4: Can we set this back to a hardcoded 6? I think it may be confusing to have this number dynamic based on word length.

src/constants/settings.ts
```diff
@@ -1,6 +1,7 @@
...
 export const ALERT_TIME_MS = 2000
 export const REVEAL_TIME_MS = 350
 export const GAME_LOST_INFO_DELAY = (MAX_WORD_LENGTH + 1) * REVEAL_TIME_MS
-export const WELCOME_INFO_MODAL_MS = 350
```
igitur, line 6: Why are you removing this line?

cwackerfuss, line 6: We'll need to add this back, just checked and it is still in use.

cwackerfuss (changes requested):

src/constants/settings.ts
igitur, line 4: Will this not set `MAX_CHALLENGES` to 8 for the default case of words of 5 letters?

cwackerfuss: We requested the same changes, hah! Great minds think alike

cwackerfuss: But +1 to this. This should just be static.

igitur: Bit of a side-note, but I started thinking about how the number of challenges relates to the word length. 

If we double the word size to 10 letters, would the user need 11 (10+1), 13 (10+3), as in this PR) or 12 (6*2) guesses? Keep in mind the other hidden factor is the alphabet size, of 26 letters, which doesn't change.

If I can find the data, I'd like to do a derivation of 3Blue1Brown's analysis and try to solve for the maximum number of guesses, given word size, while keeping alphabet and game difficulty constant.

igitur:
@cwackerfuss Either the author will have to squash the commits or you can squash and rebase.

src/constants/settings.ts
aghasemi: Hi. I agree that this is a lot language-dependent. Languages where the alphabet does not show vowels (e.g. mine, Persian), need a few more guesses for the Wordle to solve, even when most letters (i.e. consonants) are revealed.

I will just leave it a constant so that everyone can customise based on their needs.

src/constants/settings.ts
aghasemi: It was somehow absent in the new version when I created the PR :-/

aghasemi:
Hi. I have applied all changes, but there is still something red up there :D Have done something wrong?

gheja:
> export const MAX_WORD_LENGTH = solution.length

Wait what... :D It surprised me that this works, but after all, I am not a react dev.

> Hi. I have applied all changes, but there is still something red up there :D Have done something wrong?

I think @cwackerfuss has to manually approve your response to the requested changes.

cwackerfuss (approved):

cwackerfuss:
Thanks for the patience @aghasemi , just approved!

tgmgroup:
Sorry if you're busy. 

If MAX_WORD_LENGTH is now dynamic, then we only have to change the WORDS array in src/constants/wordlist.ts and VALID_GUESSES array in src/constants/validGuesses.ts? The FAQ still says:

```
FAQ
How can I change the length of a guess?

    Update the MAX_WORD_LENGTH variable in src/constants/settings.ts to the desired length.
    Update the WORDS array in src/constants/wordlist.ts to only include words of the new length.
    Update the VALID_GUESSES array in src/constants/validGuesses.ts arrays to only include words of the new length.
```

jleverenz:
Hey @tgmgroup - There's a PR with README updates at #310, but you're right, it probably should have been included with this PR.

tgmgroup:
Thanks, I didn't see the #310PR. I should have checked.

PR #269, opened 2022-02-24
Improve compatibility of date constructor

vegeta897:
Per [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters), using a string as a date constructor is strongly discouraged for compatibility reasons. Using `(year, monthIndex)` is more reliable and cleaner.


cwackerfuss:
Hi @vegeta897 - thank you! I will approve and merge this - can you run the linter with `npm run fix` (or just add a space between the args in the changed code 😄 )

vegeta897:
Whoops, sorry about that!

Thanks for accepting my PR!

PR #276, opened 2022-02-26
Wordle, but make it Starcraft

judytuna:
I can only have one fork of the original `react-wordle`, so I'm going to put my different flavors on different branches. This PR will stay open forever to track the changes from the original to make it Starcraftified!


judytuna:
Whoops, I meant to make this on my own fork of the project. Deleting this PR now.

PR #289, opened 2022-03-01
Implement Web Share API with ua-parser-js

### LINKED ISSUE #206: Share button does not work (iPhone6, Safari, iOS 12.5.5) + hints for solution

igitur:
Alternative PR to #182 that uses ua-parser-js to determine whether the user is using a mobile device instead of maintaining our own complex regex.

Fixes #206
Closes #182
Supercedes #233


cwackerfuss (approved):
LGTM! Followed along in the other PR.

PR #291, opened 2022-03-01
Took margin out of settings icon in nav bar to fix formatting. 

zbennett:


igitur:
I'm known to be quite pedantic, so excuse me. My only comment here would be that the git commit comment should be more descriptive. Something like `"Remove cog igon right margin"` would be more descriptive.

igitur:
@cwackerfuss Hope you don't mind me butting in, but unless you don't want help, I'm going to review some PRs and issues.

cwackerfuss:
Absolutely, I don't mind at all @igitur! Agreed and good catch - let's get an updated commit name, otherwise this change looks fine to me.

igitur:
@zbennett Are you comfortable with rewriting git history, and specifically editing the commit message?

zbennett:
Alright. Done.

igitur (approved):

cwackerfuss (approved):

PR #293, opened 2022-03-02
Trying to make the histogram style consistent on both ends

abrarhayat:
Trying to make the histogram style consistent on both ends. The semi-circle look looked a bit odd in my honest opinion. Would like to get some feedback.

Now

Proposed

### DIFF
--- a/src/components/stats/Progress.tsx
+++ b/src/components/stats/Progress.tsx
@@ -8,10 +8,10 @@ export const Progress = ({ index, size, label }: Props) => {
   return (
     <div className="flex justify-left m-1">
       <div className="items-center justify-center w-2">{index + 1}</div>
-      <div className="rounded-full w-full ml-2">
+      <div className="w-full ml-2">
         <div
-          style={{ width: `${5 + size}%` }}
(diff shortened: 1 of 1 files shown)


igitur:
I prefer the rounded corners, but agree that ony the left side being round is weird. Why not makes both ends round? Or only the right side rounded?

I haven't played today's Wordle, but what does that show?

abrarhayat:
The original one has rectangular edges.

igitur:
image

Looks like the spacing between bars is smaller on the original one.

And I like that it highlights the current day's bar.

cwackerfuss (approved):

cwackerfuss:
Good call - I prefer this too :)

igitur:
> image
> 
> Looks like the spacing between bars is smaller on the original one.
> 
> And I like that it highlights the current day's bar.

I'll try replicate this in a new PR.

abrarhayat:
> 

[Here](https://github.com/cwackerfuss/react-wordle/pull/295) is my solution to this.

Issue #285, opened 2022-02-27
Use of 100vh causes keyboard to be covered up on iOS webkit browsers


vegeta897:
**Overview**
The recently merged PR #209 added `h-screen` to the app container which uses `height: 100vh`. This technique for filling the screen is [known](https://stackoverflow.com/a/37113430/2612679) to cause the page to be "too tall" on iOS webkit browsers like Chrome and Safari. There are some hacky fixes out there but I don't know what approach would be best.

**To Reproduce**
1. Open reactle.vercel.app on a webkit browser (like Chrome or Safari) on iOS
2. Observe that the bottom part of the keyboard is hidden by browser UI, despite there being room above the keyboard

**Expected behavior**
The keyboard should be visible without scrolling down if there is room for it.

**Screenshots**
Chrome                                                                                                          |  Safari
:--------------------------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------------:
![image](https://user-images.githubusercontent.com/1789096/155904447-89e12e79-319c-401e-a928-27e0abf7162b.png)  |  ![image](https://user-images.githubusercontent.com/1789096/155904451-3ac27a33-d168-4d9b-b9ec-69ae4d209811.png)

**Smartphone:**
 - Device: iPhone XS Max
 - OS: iOS15.3.1
 - Browser: Safari, Chrome

diondiondion:
The simplest fix is probably just using % sizing instead of the vh unit. The downside is that `height: 100%` will have to be applied to every single ancestor of the element you want to fill the screen.

oradwell:
I have a similar issue with Firefox mobile browser. the browser UI disappears if I scroll but still annoying

TheKnowles:
@oradwell I'm collecting data points from https://github.com/cwackerfuss/react-wordle/pull/294

Could you please try this:

https://react-wordle-git-fork-theknowles-chrome-scro-8fcff7-cwackerfuss.vercel.app/

and let me know how that works.

oradwell:
@TheKnowles lovely! it is not possible to scroll anymore but the game fits perfectly into the screen. happy with this

igitur:
Fixed in #488 , #516

PR #294, opened 2022-03-02
Draft: Experiment to fix 100vh issues on mobile browsers. 

TheKnowles:
Added a custom hook that grabs the window's dimensions. On load, we then calculate the inner height of the window if we think the user is on a device smaller than a tablet. Added a little additional padding between the top navbar and the grid so that messages do not overlay the grid too much.

Tailwind was extended with a custom height that reads from this css variable.

This worked in test and on the browser emulating mobile devices. If everyone could test on additional devices, that would be very helpful.

Solution found from here: https://stackoverflow.com/questions/58886797/how-to-access-the-real-100vh-on-ios-in-css

Fixes https://github.com/cwackerfuss/react-wordle/issues/285

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -39,6 +39,7 @@ import './App.css'
 import { AlertContainer } from './components/alerts/AlertContainer'
 import { useAlert } from './context/AlertContext'
 import { Navbar } from './components/navbar/Navbar'
+import useWindowDimensions from './hooks/useWindowDimensions'
 
 function App() {
   const prefersDarkMode = window.matchMedia(
@@ -137,6 +138,17 @@ function App() {
   const clearCurrentRowClass = () => {
     setCurrentRowClass('')
   }
+
+  const { width } = useWindowDimensions()
+  // calculate actual view height (fixes chrome scroll bug re url bar)
+  useEffect(() => {
+    // If less than most tablets, set CSS var to window height.
+    let value = '100vh'
+    if (width <= 1024) {
+      value = `${window.innerHeight}px`
+    }
+    document.documentElement.style.setProperty('--real100vh', value)
+  }, [width])
 
   useEffect(() => {
     saveGameStateToLocalStorage({ guesses, solution })
@@ -241,13 +253,13 @@ function App() {
   }
 
   return (
-    <div className="h-screen flex flex-col">
+    <div className="h-dynamic flex flex-col">
       <Navbar
         setIsInfoModalOpen={setIsInfoModalOpen}
         setIsStatsModalOpen={setIsStatsModalOpen}
         setIsSettingsModalOpen={setIsSettingsModalOpen}
       />
-      <div className="pt-2 px-1 pb-8 md:max-w-7xl w-full mx-auto sm:px-6 lg:px-8 flex flex-col grow">
+      <div className="pt-8 px-1 pb-8 md:max-w-7xl w-full mx-auto sm:px-6 lg:px-8 flex flex-col grow">
         <div className="pb-6 grow">
           <Grid
             guesses={guesses}
--- a/src/hooks/useWindowDimensions.jsx
+++ b/src/hooks/useWindowDimensions.jsx
@@ -1,0 +1,26 @@
+import { useEffect, useState } from 'react'
+
+function getWindowDimensions() {
+  const { innerWidth: width, innerHeight: height } = window
+  return {
+    width,
+    height,
+  }
+}
+
+export default function useWindowDimensions() {
+  const [windowDimensions, setWindowDimensions] = useState(
+    getWindowDimensions()
+  )
+
+  useEffect(() => {
+    function handleResize() {
+      setWindowDimensions(getWindowDimensions())
+    }
+
+    window.addEventListener('resize', handleResize)
+    return () => window.removeEventListener('resize', handleResize)
+  }, [])
+
+  return windowDimensions
+}
--- a/src/index.css
+++ b/src/index.css
@@ -3,6 +3,7 @@
 @tailwind utilities;
 
 :root {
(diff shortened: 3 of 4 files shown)


TheKnowles:
Fixed lint issues.

TheKnowles:
Still needs a little work for galaxy s9+.

Actual device still has a small scroll with keyboard covered.

igitur:
Fixes issue for me on Nokia C20 + Chrome. 

@corneil Does this fix your issue with the scrolling?

igitur:
@corneil Test URL: https://react-wordle-git-fork-theknowles-chrome-scro-8fcff7-cwackerfuss.vercel.app/

TheKnowles:
iPhone 12 mini on safari has a minor amount of scrolling / keyboard overlap

image (1)

zbennett:
iPhone 12 on safari has a tiny bit of scrolling as well. 

B71FDB8F-CEAC-4752-9E23-9ADB626CFBA9

TheKnowles:
Branched off my branch to get a vercel deployment running so I can experiment some more without touching this branch.

Tried -webkit-fill-available

via https://css-tricks.com/css-fix-for-100vh-in-mobile-webkit/

Please hit this link and let me know how this works across multiple devices:

https://react-wordle-hfdw1n0c3-theknowles.vercel.app/

zbennett:
@TheKnowles It's appearing the same as above for my iphone 12. Its not a lot of scroll, just a tiny bit. It appears good on my iPhone XR which has a larger screen.

TheKnowles:
@zbennett thanks for the feedback

I also gave 

```
body {
  min-height: 100vh;
  /* mobile viewport bug fix */
  min-height: -webkit-fill-available;
}

html {
  height: -webkit-fill-available;
}
```

a spin from https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/

But the end result is about the same:

Image_20220302_171224

diondiondion:
What's wrong with just using 100% as a unit, that doesn't come with the shortcomings of the vh unit, and doesn't need JS, so should be much more robust. Just needs a few more declarations to make sure the root nodes of the app have a height declaration too.

corneil:
Looks nice on Chrome and Brave!

igitur:
> Looks nice on Chrome and Brave!

@corneil Please confirm your device.

igitur:
@TheKnowles Not sure if you're still actively working on this one. But notice that in Wordle, the guess letter boxes resize if the window size is changed. Not sure if it's capped at a max size. But this is probably relevant for your attempt to fit everything onto the screen upon load.

igitur:
@TheKnowles Anything I can do to help out with this one?

wesgurn:
Works great on iOS Safari.  LGTM!

TheKnowles:
> @TheKnowles Anything I can do to help out with this one?

@igitur This has stalled. I implemented the changes here in my downstream implementations, but there was still a little bit of scroll on iPhones w/ Safari. I don't have an iPhone and the emulators were not showing what people were seeing natively. I am okay with this mostly there solution.

I did not have any success with using 100%, etc... I am open to suggestions or other data points.

igitur:
No problem, @TheKnowles. Feedback from friends is that it still doesn't work well on an iPhone 6 SE. I think it's more a question of the small screen size, rather than the device itself. Let me get his exact screen dimensions and confirm. I can also look into this, but will need to dive very deep into TailwindCSS.

Mind giving an overview on what/why exactly you changed in this branch? And in your `vh-for-vercel` branch? Which one should we be looking at anyway? Once I know where we stand, I can try take it further.

TheKnowles:
@igitur 

This branch was the original implementation that looks at a device's dimensions and then recalculates the "real" vh. The solution here gets us about 95% of the way there. As mentioned, the only people/devices that still had problems were certain iPhones/Safari.

`vh-for-vercel ` was/is an experimental branch to avoid touching this branch during PR review.

I tried a handful of various CSS tricks I found on the internet in an attempt to get Safari/iPhone to be consistent with other browsers/devices. None of them were particularly helpful and didn't do much better then what I have here in this branch.

I would look at this branch.

Bearborg:
I ran into this issue on my own fork of this project, and got curious about what the _original_ Wordle is doing. It turns out they simply set `height: 100%;` on the `<html>` element, the `<body>` element, etc. all the way down to the main container div. I applied the same styling to my own fork [here](https://github.com/Bearborg/chozordle/commit/b0cff686672f685b45939f8806cff255c1995cc5) (ignore the change to index.css, that was a previous failed solution), and the [results](https://chozordle.vercel.app/) look pretty good to me on an iPhone 13; I think this may be a way to achieve the same results as your PR using pure CSS.

igitur:
> I ran into this issue on my own fork of this project, and got curious about what the _original_ Wordle is doing. It turns out they simply set `height: 100%;` on the `<html>` element, the `<body>` element, etc. all the way down to the main container div. I applied the same styling to my own fork [here](https://github.com/Bearborg/chozordle/commit/b0cff686672f685b45939f8806cff255c1995cc5) (ignore the change to index.css, that was a previous failed solution), and the [results](https://chozordle.vercel.app/) look pretty good to me on an iPhone 13; I think this may be a way to achieve the same results as your PR using pure CSS.

@Bearborg This looks promising. Would you mind submitting those specific changes as a PR?

On an iPhone 8 I still see that one has to scroll to see the entire content:
image

But I'm starting to think that's a separate issue where the tiles have to change size if the screen size is too small. 

Let's get that PR first, please.

igitur:
Fixed in #488 , #516 .

Can I close this?

TheKnowles:
Yes, thank you.

PR #300, opened 2022-03-04
Add debug option to show solution

igitur:
While reviewing PRs and debugging issues, I often need to know the answer upfront. I know I can get it by inspecting local storage, but that's effort.

This PR (still work in progress) adds support for an environment variable to optionally show the solution above the grid.


igitur:
Your comments please @gheja @TheKnowles @cwackerfuss.

TheKnowles:
+1 This would be a useful feature.

Can .env be hot reloaded or do you have to bring the server down/up again?

gheja:
Maybe the best option would be to have a "cheats enabled" flag in .env that would only add a button on the navbar that pops up a modal like Settings but with debug/cheat options like this.

src/constants/settings.ts
```diff
@@ -4,3 +4,7 @@ export const ALERT_TIME_MS = 2000
 export const REVEAL_TIME_MS = 350
 export const GAME_LOST_INFO_DELAY = (MAX_WORD_LENGTH + 1) * REVEAL_TIME_MS
 export const WELCOME_INFO_MODAL_MS = 350
+
+export const DEBUG_SHOW_SOLUTION =
+  (process.env.REACT_APP_DEBUG_SHOW_SOLUTION?.toLowerCase() ?? 'false') ===
+  'true'
```
cwackerfuss, line 10: Isn't this logic the same as:
```
process.env.REACT_APP_DEBUG_SHOW_SOLUTION?.toLowerCase() === 'true'
```

src/components/Solution.tsx
```diff
@@ -0,0 +1,21 @@
+type Props = {
+  isVisible: boolean
+  solution: string
+}
+
+export const Solution = ({ isVisible, solution }: Props) => {
```
cwackerfuss, line 6: Can we call this `DebugSolution` so it's super clear it's not used in feature code?

src/App.tsx
```diff
@@ -248,6 +250,7 @@ function App() {
         setIsSettingsModalOpen={setIsSettingsModalOpen}
       />
       <div className="pt-2 px-1 pb-8 md:max-w-7xl w-full mx-auto sm:px-6 lg:px-8 flex flex-col grow">
+        <Solution isVisible={DEBUG_SHOW_SOLUTION} solution={solution} />
```
cwackerfuss, line 253: Nit: I prefer this since it makes the `Solution` more single responsibility.
```suggestion
        {DEBUG_SHOW_SOLUTION && <Solution solution={solution} />}
```

src/constants/settings.ts
igitur: Yes, this is me still learning TypeScript. I'll simplify this.

src/components/Solution.tsx
igitur: Sure.

src/App.tsx
igitur: Ok, again, me learning TypeScript / React.

igitur:
> Can .env be hot reloaded or do you have to bring the server down/up again?

Not sure about .env, will check, but you can hot reload by setting `export const DEBUG_SHOW_SOLUTION = true` in settings.ts

gheja:
> my goal is to minimise the number of clicks to see the solution

In that case a dedicated button would be the best. I suggested the extra modal to have the debug version as close as possible to prod version.

> what other kinds of cheats would you enable?

I am not sure yet. Maybe a "clear local storage" or a "pick a new word to guess", I'm not sure, anything that helps the development or bug hunting but not part of the game. If there's a way to decide during build time to exclude the debug parts altogether, that would be useful too, and we'd have only one or two big chunks of debug code instead of scattered through the code.

cwackerfuss (approved):

cwackerfuss:
I think this is ready to go -- is this intended to still be in draft mode @igitur ?

igitur:
@cwackerfuss Welcome back!

Yes, still meant to be in draft mode. I want to incorporate the feedback. I just kind of forgot about this PR.

PR #305, opened 2022-03-08
Fix max challenge var

czhou23:


igitur:
This looks like an accidental PR.

Issue #302, opened 2022-03-07
unit tests broken: ReferenceError: Cannot access 'mockSolutionGetter' before initialization


jleverenz:
Two PRs went in recently that are not compatible.

* #252 introduced a module level dependency on `solution`
* #297 introduced new unit tests that attempt to mock `solution` out

From what I can tell tests on each PR passed, it's just once combined, the following error is generated:

```
FAIL src/lib/share.test.ts
  ● Test suite failed to run

    ReferenceError: Cannot access 'mockSolutionGetter' before initialization
```

Creating this issue to help track, I can propose a workaround soon.

denik1981:
Can someone please review the PR(#330) I've submitted a few days ago? 
It should fix this testing issue.

PR #308, opened 2022-03-08
Add a workaround for unit test ReferenceError

jleverenz:
Addresses #302. This feels a little hacky to me, looking for input. Another option would be a larger refactor of how `solution` is used across the app, or maybe some jest mocking fu beyond my knowledge.


jleverenz:
Alternative refactor proposed in #310. Only one PR is needed, welcome any feedback/suggestions.

igitur:
Not that I have a better suggestion, but, yes, this is indeed very hacky...

PR #310, opened 2022-03-09
Refactor word/solution length to be variable

jleverenz:
Addresses #302 by refactoring MAX_WORD_LENGTH to be dynamic. Instead of including it in `constants` this refactor changes the app to use `solution.length` where needed.

Since the app now supports flexible, varied length solutions and guesses per day (#252), it seemed moving related variables out of `constants` to be a reasonable refactor.

There's an alternative workaround in #308. Only one is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,19 @@ _Want to add one to the list? Just make a pull request or [let us know via a com
 
 ### How can I change the length of a guess?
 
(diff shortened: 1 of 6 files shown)


cwackerfuss (approved):

cwackerfuss:
Awesome,  this looks great.

muhelen:
Hi, does anyone knows how to include long sentence words with spaces?
Ex. Never Say Never Again

PR #322, opened 2022-03-14
Fix getWordOfDay for DST-affected timezones

### LINKED ISSUE #321: Daylight Savings Time shifted next word start by an hour
**Describe the bug**
We just had Daylight Savings Time in the US, and the next puzzle is off by an hour (1am instead of 12am).

**To Reproduce**
This one may be a bit tricky to reproduce... you'll have to mock the time during a daylight savings time shift...

**Expected behavior**
The next word should be available at midnight. Instead it's available at 1am after the "spring forward" daylight savings time shift.

**Desktop (please complete the following information):**
 - OS: macOS 12.2.1 
 - Browser: Chrome
 - Version: 99.0.4844.51

vegeta897:
Fixes #321 

**Old method:** Get the quantity of 86400000-millisecond chunks between the epoch and the current time

The issue with this is that if the epoch is not during DST and the current date is (or vice versa) then the word no longer resets at midnight. It will be off by one hour until DST ends (or begins, if the epoch was during DST)

**New method:** Javascript doesn't have a simple method for getting the number of days between two days, but it does let you add days to a date. By running a simple loop, we add days to the epoch until it equals today's date. I found this idea [here](https://flaviocopes.com/how-to-count-days-between-dates-javascript/).


ThatOneCalculator (approved):

igitur:
How about relying on a 3rd party library to manage this. From what I gather, DST isn't the only problem with diffing dates, although it's the most common. Haven't tested it, but here's something promising: https://github.com/kshetline/tubular_time

vegeta897:
> How about relying on a 3rd party library to manage this. From what I gather, DST isn't the only problem with diffing dates, although it's the most common. Haven't tested it, but here's something promising: https://github.com/kshetline/tubular_time

Perhaps, but that seems like overkill (especially that particular library's filesize) for what's needed here. The only real expectation that needs to be met is that the next day is available at midnight. The proposed changes take care of that. But I'd be curious to hear about a specific scenario that doesn't work with this new code.

farrukhshahzad:
will it work if DST in not one full hour?

vegeta897:
> will it work if DST is not one full hour?

The code doesn't care about hours/minutes/seconds, it just gets 0:00:00.000 of the current day, and adds calendar days to the epoch until that's met. DST shouldn't affect it in any way.

farrukhshahzad:
> > will it work if DST is not one full hour?
> 
> The code doesn't care about hours/minutes/seconds, it just gets 0:00:00.000 of the current day, and adds calendar days to the epoch until that's met. DST shouldn't affect it in any way.

Thanks. I have modified to have 2 words a day. So I just divide `const msInDay = 86400000` by 2 to give us a new word every 12 hours.

praneeth-m:
Does anyone know what's holding this PR from being approved/merged?

cwackerfuss (approved):

vegeta897:
Thanks for accepting my PR!

PR #330, opened 2022-03-17
Fix unable to mock solution in shared tests

### LINKED ISSUE #302: unit tests broken: ReferenceError: Cannot access 'mockSolutionGetter' before initialization

denik1981:
### Summary of changes
This PR tries to solve the problem of the solution not being mocked properly in the shared tests. 
This committed change will assure that './words' will be imported dynamically only when needed and after the mock return value has been resolved in order to be correctly applied to the getter that is defined. Once './words' is imported into the running module, `solution` will have value in mockSolutionGetter.mockReturnValue to resolve and let its getter act upon it.   

To allow typescript isolated modules feature keep working, an empty export at the very end was needed as well now that there are no static imports in the entire module.

I have also renamed the second test to adhere to the `eslint` rule of avoiding using the same name on different running tests

Fixes #302

### How to test this?

- `npm run test`

### DIFF
--- a/src/lib/share.test.ts
+++ b/src/lib/share.test.ts
@@ -1,5 +1,3 @@
-import { generateEmojiGrid } from './share'
-
 const mockSolutionGetter = jest.fn()
 jest.mock('./words', () => ({
   ...jest.requireActual('./words'),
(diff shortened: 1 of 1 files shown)


src/lib/share.test.ts
```diff
@@ -9,26 +7,28 @@ jest.mock('./words', () => ({
...
     mockSolutionGetter.mockReturnValue('ABCDE')
 
-    const grid = generateEmojiGrid(guesses, tiles)
+    const grid = (await import('./share')).generateEmojiGrid(guesses, tiles)
     const gridParts = grid.split('\n')
     expect(gridParts[0]).toBe('PPCPP')
     expect(gridParts[1]).toBe('AAAAA')
     expect(gridParts[2]).toBe('CCCCC')
   })
-  test('generates grid for ascii', () => {
```
jleverenz, line 23: Thanks for cleaning up the name!

src/lib/share.test.ts
```diff
@@ -9,26 +7,28 @@ jest.mock('./words', () => ({
...
-    const grid = generateEmojiGrid(guesses, tiles)
+    const grid = (await import('./share')).generateEmojiGrid(guesses, tiles)
     const gridParts = grid.split('\n')
     expect(gridParts[0]).toBe('PPCPP')
     expect(gridParts[1]).toBe('AAAAA')
     expect(gridParts[2]).toBe('CCCCC')
   })
 })
+
+export {}
```
jleverenz, line 34: Just a question, as I'm not too familiar with this pattern - is the export needed for anything? `npm test` seemed to pass for me without.

denik1981: Once you remove all the static imports, the so-called "module" becomes "non-module" for the typescript compiler unless you add, at least, one export. If you don't have anything to export, you need to create an empty export. This is required when using `isolatedmodules` in  `tsconfig,json`.

**Without `export {}`**
image

igitur:
I believe #379 is a simpler solution to this.

denik1981:
> I believe #379 is a simpler solution to this.

It is a very subjective statement. Depending on what you mean by simple. 
This PR changes only one file with +6 insertions and -6 deletions.

#379 changes 9 files, having +69 insertions and -20 deletions

igitur:
The current unit tests attempt to mock a partial. That's a much more complicated concept to grok than just injecting the required values via a parameter. I don't think this is even a gray area.

PR #336, opened 2022-03-19
Support self-hosted Plausible Analytics

vegeta897:
This PR adds a new environement variable, `REACT_APP_PLAUSIBLE_HOST` which allows the use of a [self-hosted](https://plausible.io/docs/self-hosting) instance of Plausible.

* A bullet-point was added to the README explaining this optional variable
* `index.html` was updated to use the custom host if not null

Tested & working on my own self-hosted Plausible instance:
image


cwackerfuss (changes requested):
Rather than creating a second env variable, can we repurpose the existing variable so the entire URL is included rather than just the base?

vegeta897:
> Rather than creating a second env variable, can we repurpose the existing variable so the entire URL is included rather than just the base?

Unfortunately that would just make setup more complicated. Keep in mind that the `REACT_APP_PLAUSIBLE_DOMAIN` is the site being tracked, and `REACT_APP_PLAUSIBLE_HOST` is the site providing the tracking. They are not necessarily related, and they don't even need to be on the same domain. Likely they won't be, if the user already has Plausible self-hosted before installing this project.

The script URL points to an automatically created js file that lives on your self-hosted Plausible URL. So if I'm hosting it at `plausible.example.app`, my script would automatically be available at `plausible.example.app/js/plausible.js`. The only way the user's Reactle URL and Plausible URL would overlap in a way that let us configure only one URL is if the user manually copied that .js file into their Reactle project. And they'd have to remember to update it when updating their Plausible instance. That just seems like more work than merely defining a second env variable, and increases the chance of user error.

pietroppeter:
fwiw (I did implement the original plausible support) I agree with @vegeta897 that two variables seem to be the best approach as they serve different purposes.

pietroppeter (approved):
added just a suggestion on how to rephrase note in README, but from what I can tell it looks good :)

igitur:
Please resolve the conflicts and I'll look at possibly merging the PR.

vegeta897:
Conflict resolved, thanks.

igitur:
Github still reports conflicts. 

And please whitelist the dummy plausible domains in the linkcheck config file. See CI task output for details.

vegeta897:
I've whitelisted the dummy URL, but there are still 2 links unrelated to this PR that aren't resolving.

Github is telling me there are no conflicts.

vegeta897:
I just received a DMCA notice from Github, targeted at my fork of this repository which is the basis of this PR 😂

So I guess that's that!

PR #345, opened 2022-03-22
Histogram fixes

igitur:
When the game is lost, none of the histogram bars should be highlighted. Rather they should all show in gray.

When no games have been won every, we currently have a division by zero, resulting in the histogram bars showing at "100%". 

This PR solves these issues.


src/components/stats/Histogram.tsx
```diff
@@ -3,20 +3,37 @@ import { Progress } from './Progress'
 
 type Props = {
   gameStats: GameStats
+  isGameWon: boolean
   numberOfGuessesMade: number
 }
 
-export const Histogram = ({ gameStats, numberOfGuessesMade }: Props) => {
+const isCurrentDayStatRow = (
```
abrarhayat, line 10: I think a separate function for this purpose is unnecessary. The fix to your comment [here](https://github.com/cwackerfuss/react-wordle/pull/304) is quite the same right?

src/components/stats/Histogram.tsx
```diff
@@ -3,20 +3,37 @@ import { Progress } from './Progress'
...
+  const maxValue = Math.max(...winDistribution, 1)
 
   return (
     <div className="columns-1 justify-left m-2 text-sm dark:text-white">
       {winDistribution.map((value, i) => (
         <Progress
           key={i}
           index={i}
-          currentDayStatRow={numberOfGuessesMade === i + 1}
+          isCurrentDayStatRow={isCurrentDayStatRow(
```
abrarhayat, line 32: I think this is better off with `numberOfGuessesMade === i + 1 && isGameWon`

src/components/stats/Histogram.tsx
```diff
@@ -3,20 +3,37 @@ import { Progress } from './Progress'
...
+}
+
+export const Histogram = ({
+  gameStats,
+  isGameWon,
+  numberOfGuessesMade,
+}: Props) => {
   const winDistribution = gameStats.winDistribution
-  const maxValue = Math.max(...winDistribution)
+  const maxValue = Math.max(...winDistribution, 1)
```
abrarhayat, line 24: Yes this fix was necessary. The `size` property in the `Progress` component becomes `NaN` when we have no wins at all.

cwackerfuss (approved):
LGTM!

PR #349, opened 2022-03-23
Linkcheck

modem7:
To make sure that all the links within the readme are live, not duplicates, and prevent dead links in future.

[Example](https://github.com/petersem/dockerholics/runs/5843557873?check_suite_focus=true#step:4:1)

This would make the "`Projects built using this repo`" section easier to manage when PR's are submitted.

Edit: This was resubmitted as I erroneously deleted my previous repo which contained https://github.com/cwackerfuss/react-wordle/pull/333 which caused it to close.


modem7:
Updated example as logs had expired

gheja:
I like the idea of validating the links but I don't really understand how this would work.

I think this should not be a part of the standard build/test process or the PR workflow as giving a hit to each webpage every time a build runs on any machine sounds excessive. And if it would be a part of the workflow it could be abused by putting links in the readme with malicious intent.

Also, this would bring more external dependencies, increasing the chance of a failed build with every new link, especially in case of unreliable networks or temporary problems with any of the sites.

I think a bot that periodically checks the links and opens pull requests with the failed checks would be really nice to have.

Anyway, this is just my opinion, based on a thing that I don't fully understand, luckily I am not the one who makes the decisions :)

modem7:
> I like the idea of validating the links but I don't really understand how this would work.
> 
> I think this should not be a part of the standard build/test process or the PR workflow as giving a hit to each webpage every time a build runs on any machine sounds excessive. And if it would be a part of the workflow it could be abused by putting links in the readme with malicious intent.
> 
> Also, this would bring more external dependencies, increasing the chance of a failed build with every new link, especially in case of unreliable networks or temporary problems with any of the sites.
> 
> I think a bot that periodically checks the links and opens pull requests with the failed checks would be really nice to have.
> 
> Anyway, this is just my opinion, based on a thing that I don't fully understand, luckily I am not the one who makes the decisions :)

In this particular workflow, it would be limited purely to the README file, and wouldn't affect the general build of the app itself. 

So effectively, when someone adds a new link to the readme, it would run the workflow, and fail if the links are invalid/dead, making it easier to see the who/what/where/why.

gheja:
> In this particular workflow, it would be limited purely to the README file, and wouldn't affect the general build of the app itself. It would only run if there were a change to the readme file.

That sounds great, thanks for explaining!

> So effectively, when someone adds a new link to the readme, it would run the workflow, and fail if the links are invalid/dead, making it easier to see the who/what/where/why.

If I get it right (from the logs you provided) it would validate all the links. Would it block the PR if it found a broken link or just an throw a warning? For example, if someone wanted to add a link but in the meantime another one gets broken, would they need to fix (remove) the broken link to have their PR merged?

I don't want to nitpick I just would like to understand it. It actually sounds like a great tool and I'm curious.

igitur (approved):

modem7:
@cwackerfuss - any thoughts regarding this PR?

Cheers!

PR #359, opened 2022-03-31
Add Jwordle (Japanese version) to README.md

woelpad:


igitur:
I think it's time to split out all the downstream implementations into a IMPLEMENTATIONS.md. The README.md is getting a bit full. And to be honest, I don't think many people care anymore about seeing a list of implementations. It was a novel think in Jan 2022, but not that interesting anymore. And I say that, having added my own implementation too.

woelpad:
Well, this Temblung thing is definitely not Japanese, and as you confirmed
for yourself, it pretty much could only be Javanese. A simple typo, a
simple fix (I hope).

Op do 31 mrt. 2022 om 17:32 schreef Gabor Heja ***@***.***>:

gheja:
I think we should keep at least a select dozen or so forks in the README.md and a link to the other .md file, as it looks much better to a first time visitor, seeing the diversity, flexibility and potential of the project directly in the README is a huge plus. IMHO.

woelpad:
The proposed split-up of README.md might be a valid solution, but would it not make better sense to first merge all the current changes to the file before attempting to split it up? This particular pull request still needs approval. Please move ahead.

cwackerfuss:
+1 to what @gheja suggested. I don't particularly feel the list has grown so long that it's detracting from the README. I do think it's possible the FAQs are more important than the project list, so maybe the solution is to put the project list at the bottom after all documentation?

woelpad:
That would work for me.

On another note, I notice that the most successful Japanese Wordle, which
is also a clone of this project, is not in the list. Am I allowed to add it
or should I request the author to do it himself?

https://taximanli.github.io/kotobade-asobou/

PR #360, opened 2022-03-31
Resolves #266 - compressing wordlist and integrating with module 

### LINKED ISSUE #266: Reduce download size by compressing wordlist
**Is your feature request related to a problem? Please describe.**
The main js bundle is around 360kb in size, which is significant, and a major factor in my decision not to use this version for Byrdle: I have ~50k users/day, meaning the daily bandwidth would be around 17 GB/day.   

**Describe the solution you'd like**
100kb of the js bundle is the list of permitted guesses, but you can (at least) halve that using a few simple tricks:
- Store the list as a single string (removes quotation marks and commas: 37.5% saving) and chop into 5-letter chunks after deployment
- Divide the list into 26 arrays, based on starting letter, and remove the starting letter (another 20% saving, after doing the first suggestion).

**Describe alternatives you've considered**
You could go further by changing encodings, but the above (which I use in my [deployed code](https://github.com/rbrignall/open-byrdle/blob/main/src/words_5_6.ts)) doesn't have a major impact on serving the page. 

Another option would be to use a Trie. I havne't investigated that.

**Additional context**
Here is the (Svelte) [wordlist](https://github.com/rbrignall/open-byrdle/blob/main/src/words_5_6.ts) that I use, including the code that recovers the array of words. As you can see, I can deploy 5- and 6-letter words in less than the original 5-letter array...

ryanlovallo:
Resolves #266  

Changed the format for storing the words in the VALID_GUESSES variable (src/constants/validGuesses.ts), reducing the size of the file from 139kb to 52kb. Changes in src/lib/words.ts indicate the new method of looking up words in the dictionary (comments elaborate on the exact procedure).


igitur:
> I think this compression should be done as part of a build/deploy process, not in the source code. The source should be as easily maintainable as possible.

Definitely. This PR will also conflict with other PRs that require words of different lengths. Please abandon this PR.

PR #364, opened 2022-04-01
Converted the modal close div to a button.

Bracciata:
This makes the app more accessible and allows keyboard only users to close out of the modal easily.

### DIFF
--- a/src/components/modals/BaseModal.tsx
+++ b/src/components/modals/BaseModal.tsx
@@ -47,10 +47,10 @@ export const BaseModal = ({ title, children, isOpen, handleClose }: Props) => {
             leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
           >
             <div className="inline-block align-bottom bg-white rounded-lg px-4 pt-5 pb-4 text-left overflow-hidden shadow-xl transform transition-all sm:my-8 sm:align-middle sm:max-w-sm sm:w-full sm:p-6 dark:bg-gray-800">
-              <div className="absolute right-4 top-4">
+              <div onClick={() => handleClose()} role="button" tabIndex={0} aria-pressed="false" className="absolute right-4 top-4">
                 <XCircleIcon
                   className="h-6 w-6 cursor-pointer dark:stroke-white"
-                  onClick={() => handleClose()}
+                  
                 />
               </div>
               <div>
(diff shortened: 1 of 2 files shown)


igitur:
Please post a screenshot of the new version.

cwackerfuss (dismissed):

igitur:
Welcome back, @cwackerfuss . If you need more help with all the PRs and issues in some way, let me know.

Bracciata:
Accidentally added the commits from the other a11y pr.

gheja:
Now the Close button is the first element in the modal, this means it gets focused when the modal opens. It looks much better for the Info modal where the link at the bottom was focused by default.

One note however: in the Vercel preview I cannot close the modal using the keyboard, Enter tries to submit the guess, Space gets ignored. The link opens just fine on Enter. (Windows 10 + Chrome 99)

src/components/modals/BaseModal.tsx
```diff
@@ -47,10 +47,10 @@ export const BaseModal = ({ title, children, isOpen, handleClose }: Props) => {
             leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
           >
             <div className="inline-block align-bottom bg-white rounded-lg px-4 pt-5 pb-4 text-left overflow-hidden shadow-xl transform transition-all sm:my-8 sm:align-middle sm:max-w-sm sm:w-full sm:p-6 dark:bg-gray-800">
-              <div className="absolute right-4 top-4">
+              <div onClick={() => handleClose()} role="button" tabIndex={0} aria-pressed="false" className="absolute right-4 top-4">
```
igitur, line 50: This doesn't handle the keyboard presses as I expected it would. Should this not be converted to a real `<button>` tag? e.g.

```
<button
  onClick={() => handleClose()}
  tabIndex={999}
  aria-pressed="false"
  className="absolute right-4 top-4"
>
    <XCircleIcon className="h-6 w-6 cursor-pointer dark:stroke-white" />
</button>
```

igitur: Also remember to run prettier on your code to conform to style.

igitur: cc @gheja

igitur:
Note the failing test because of the malformatted code. https://github.com/cwackerfuss/react-wordle/runs/6271545848?check_suite_focus=true

PR #365, opened 2022-04-01
Added Aria Label to the keys in the keyboard.

Bracciata:
This gives users of screen readers a chance to know if the letters are in the word, if they are present, and if they are in the correct position. In the below screenshots without this feature it would just say the letter instead of status too to a screen reader.
Screen Shot 2022-04-01 at 12 25 56 PM 1
Screen Shot 2022-04-01 at 12 24 57 PM
Screen Shot 2022-04-01 at 12 24 28 PM
Screen Shot 2022-04-01 at 12 25 27 PM

Accessibility Fix.

### DIFF
--- a/src/components/keyboard/Key.tsx
+++ b/src/components/keyboard/Key.tsx
@@ -54,7 +54,7 @@ export const Key = ({
   }
 
   return (
-    <button style={styles} className={classes} onClick={handleClick}>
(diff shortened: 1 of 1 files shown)


igitur:
Cool idea, but I think this is outside the scope of being a Wordle clone.

igitur (approved):

PR #372, opened 2022-04-06
Show share failure notification

### LINKED ISSUE #371: Add a notification about failing to properly execute Share
The Share feature only works when the page is loaded over HTTPS.

See these issues:
https://github.com/cwackerfuss/react-wordle/issues/331
https://github.com/cwackerfuss/react-wordle/issues/367

The Share button now silently ignores the failure of copying to clipboard.

It might fail silently under other conditions as well.

I believe it would be the best to give a notification to the player about this failure. (This would be more robust than trying to detect if the requirements are met before pressing the Share button, IMHO.)

igitur:
Fixes #371


igitur:
cc @gheja

gheja:
I think that looks great. Thank you!

src/lib/share.ts
```diff
@@ -41,9 +42,19 @@ export const shareStatus = (
...
+        navigator.clipboard
+          .writeText(textToShare)
+          .then(handleShareToClipboard)
+          .catch(handleShareFailure)
+      } else {
+        handleShareFailure()
+      }
+    }
+  } catch (error) {
+    handleShareFailure()
```
cwackerfuss, line 57: Is this try/catch block still necessary with `handleShareFailure` also being caught on L51?

igitur: @cwackerfuss I'll look into it soon.

igitur: OK, so while testing on different browsers, I realised they don't all support exactly the same API. So rather than checking whether every every property and method is available, I just wrapped the entire block in a try/catch. For example (and this is just hypothetical), if a browser doesn't have `navigator.clipboard.writeText`, then an exception would occur, but I don't think the `.catch` on L51 will catch it. So I think we should keep this as a failsafe.

igitur:
This is still draft, because I want to add a new modal that is displayed when the Web Share API and clipboard methods fail. The user will then be able to manually copy the share text from a textarea element.

igitur:
On 2nd thought, I'll log a separate PR for that. Let's get this through and out in the wild first.

gheja:
Looks great, just one thing: the message disappears a bit too soon, I had to press the button 3 times to read the whole message. But of course, the player can do the same.

To emulate the failure I just did this in the console:
```javascript
navigator.clipboard.writeText = null
```

igitur:
> the message disappears a bit too soon

I added a new `LONG_ALERT_TIME_MS = 10000` variable in `settings.ts` and implemented this for this notification.

PR #373, opened 2022-04-08
Update README.md

aghasemi:
Add Reverdle and RareWordle


aghasemi:
Hi. Any idea why a test fails here? The only thing I have changed is the README 😄

Issue #261, opened 2022-02-23
Share Button issues within Facebook mobile browser


joedel263:
A friend just pinged me because the "share button isn't working" though the shared to clipboard message shows up.

We backtracked it and found that she was only having the issue (samsung galaxy phone) when playing the game within the browser that is built in to facebook.

can confirm same issue on my phone (pixel 5) that when you play within the facebook browser and hit the share button, it says it's copying the text, but it does not.

The demo linked in https://github.com/cwackerfuss/react-wordle/pull/182 (Web Share API) produces same result on both devices.  the "copied to clipboard" message is shown but text doesn't copy.  (hitting open in browser and sharing worked properly giving the different share options)

I haven't played with facebook in forever but it wouldn't surprise me in the least if this was something they did on purpose to force sharing within their environment..

gheja:
Thanks for reporting and pointing out this issue!

From games point of view Facebook/Messenger built-in browser is a curse... IIRC it does not store cookies and local storage (or deletes after closing it) and is severely limited.

It might be a good idea to detect it and give the user a warning that they should open it in a proper browser otherwise they will lose their progress.

There is a [detect-inapp](https://www.npmjs.com/package/detect-inapp) package which does not look up-to-date at all, the [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) looks more promising as it has a possible "Facebook" value for `browser.name` and it is being used in Pull Request #233. Although I am just guessing here, not sure which user agent gets categorized as "Facebook", this needs throghout testing.

PR #374, opened 2022-04-10
Discourage use of in-app browsers

igitur:
Discourage the use of in-app browsers (e.g. what happens when you click a link on a default Facebook app installation)


gheja:
Related Issues: #233, #261

cwackerfuss (approved):
Thanks @igitur !

PR #379, opened 2022-04-17
Fix unit tests

### LINKED ISSUE #302: unit tests broken: ReferenceError: Cannot access 'mockSolutionGetter' before initialization

igitur:
Instead of mocking the `words` partial, I rather use parameter injection to inject the solution into the relevant methods.

I believe this is a better solution than #330 as this PR enforced better inversion of control and simplifies the tests.

Fixes #302 
Closes #330


oradwell (approved):
looks good to me. merged into [jazle](https://github.com/oradwell/jazle) in [#62](https://github.com/oradwell/jazle/pull/62)

cwackerfuss (approved):
This looks good to me too. Thanks @igitur -- by the way, I sent you a collaborator request! You've stuck with this even while I've been distracted with starting a new job and life, and I really appreciate that.

PR #393, opened 2022-04-23
Implement different messages when problem was solved on last guess

igitur:
Display a random string from a different list of messages if the problem was solved on the last guess.

This is something the original also does.


igitur:
@gheja Mind giving some feedback on this one?

PR #418, opened 2022-05-11
Add support for migrating stats to new devices

### LINKED ISSUE #417: Migrate statistics to new device
When users change browser or devices, they lose their stats / streaks / game state. This happens in the original Wordle too. But I thought it would be a useful feature to be able to migrate one's stats to a new device. This is a departure from our goal of being a faithful Wordle clone, but I think it's worth it to at least consider it.

I've done a proof of concept on the test version of my downstream implementation at https://wortel.netlify.app . Sorry, it's not in English, but Chrome's autotranslate will get you far. 

Click on the Stats icon and you'll see a new button that leads you to the migration dialog:
image

The idea is to generate a base64 encoded string of the stats that one can copy on the old device and paste into the new device. It's not meant to be super secure. You could theoretically hand out your code to anyone and they can overwrite their own stats with what they receive.

From old device:
image

On new device (using invalid code):
image

On new device (using valid code), notice confirmation prompt at top:
image

Please give some feedback, on whether this should be submitted as a PR on this repo (else I'll leave it downstream) and any other feedback.

cc @gheja @TheKnowles @cwackerfuss

igitur:
Experimental PR to enable migration of statistics and game state to other devices. Feedback welcome.

The migration code is generated by:
- Apply Blowfish encryption to the JSON representation of the statistics and game state
- Apply Base64 encoding

Closes #417


igitur:
image
image
image

igitur:
The feedback I'm receiving is that this works well. I'll merge and we can refine if required.

PR #451, opened 2022-06-19
Refactor words.ts and add unit tests

igitur:
Refactor words.ts and add unit tests


igitur:
This would add support for e.g. weekly challenges, instead of daily.

Issue #465, opened 2022-07-12
The word "faire" is in wordlist.ts but not in validGuesses.ts


discapes:
The word faire has a chance to be an answer to a wordle, but since it's not a valid guess, it's impossible to get right. This might not be the only one. Affects all platforms.

tejasraman:
Maybe file a PR?

igitur:
Yes, please submit a PR with all missing English words. 😂

tejasraman:
I will do it later today... I already have the valid guesses list used in NYT's Wordle, maybe that would work?

PR #483, opened 2022-08-05
Added new words to valid guesses - #465

tejasraman:
Issue #465 states that some words are in the answer list but not the valid guesses.  I compiled a list based on [https://github.com/dwyl/english-words](https://github.com/dwyl/english-words) and added it to validGuesses.ts.


igitur:
OK, but you're also removing many words in this PR.

tejasraman:
I didn't see that. I will merge this and the existing list, sorry. 
AFAIK, that word list should contain most English words.....

I'll close the PR and resubmit with a proper list.

igitur:
I think it's worth saying that this project is meant to use as a base for your own downstream implementation. So I don't expect the list of English words in this repo to be complete (if that definition even exists). That said, I'm not opposed to adding missing words.

PR #485, opened 2022-08-06
Including a portuguese option

duarch:
https://www.persi.ga/ is a 7 characters portuguese translation. Thank you for the great job you have done!

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,6 @@ For mobile and wearable devices and smart TVs, sharing of results is initially a
 
 ### Fun themes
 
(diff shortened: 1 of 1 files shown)


igitur:
As per https://github.com/cwackerfuss/react-wordle/pull/349#issuecomment-1119495387 please remove the invalid current links. Please check them all for false positives.

duarch:
@igitur I remove the https://tirea.learnnavi.org/wordle , but this other https://chipotlele.herokuapp.com/ it is unstable sometimes its on (200) sometimes is off. 

Could you please consider it an exception?

igitur:
Yes, I agree

PR #488, opened 2022-08-09
Make layout responsive to height

igitur:
Addresses some of the issues in #429.

Add new custom media query with `min-height`. See https://tailwindcss.com/docs/screens#custom-media-queries

Cells and Keys now smaller and grow only once the min-height has been reached. Also some tweaks to padding. 

I'll need feedback from iPhone users.


stenkast:
I've got an iPhone 7 and decided to go ahead and test it. 

[Safari](https://i.imgur.com/0XpJ3er.jpg)

[Firefox](https://i.imgur.com/5wMOymv.jpg)

It seems like it's a perfect solution for Firefox and Chromium, but doesn't work as intended on Safari. I still think this is a better solution than #429 because that one always broke for me.

igitur:
@stenkast Would you perhaps be open to chatting on WhatsApp or Telegram to debug this easier? If so, email me on my same username @ gmail.

stenkast:
@igitur I did some googling and found out that this [npm package](https://www.npmjs.com/package/react-div-100vh) exists and solves the issue with Safari cutting it off. It also removes the need to add full height in the index.html file as well.

I just wrapped everything in App.tsx. It might be something to consider?

igitur:
Thanks @stenkast . The "Safari cutting off" issue is a separate problem, because of how Safari defines the viewport. See https://medium.com/@susiekim9/how-to-compensate-for-the-ios-viewport-unit-bug-46e78d54af0d

This PR just tries to address the issue of fitting all the content into whatever is deemed the viewport. On small devices (iPhone or not) the current implementation renders too large elements. And this tries to fix it.

stenkast:
> Thanks @stenkast . The "Safari cutting off" issue is a separate problem, because of how Safari defines the viewport. See https://medium.com/@susiekim9/how-to-compensate-for-the-ios-viewport-unit-bug-46e78d54af0d
> 
> This PR just tries to address the issue of fitting all the content into whatever is deemed the viewport. On small devices (iPhone or not) the current implementation renders too large elements. And this tries to fix it.

That's the very article I read, that pointed me to the solution that I provided. It's the "Solution 3" in that article and it solved the issue for me. Safari no longer cuts the keyboard off, and the grid is appropriately sized.

igitur:
Yes, but it's not relevant to THIS PR.

stenkast:
> Yes, but it's not relevant to THIS PR.

How is it not relevant, the title literally states "make layout responsive to height" and that npm package does exactly that by calculating the correct HEIGHT? But [alright](https://i.imgur.com/o52gSwC.png).

igitur:
For example, I know of a user whose viewport visible viewport dimensions are 320x468 and this repo tries to squash the content in for such small devices.

stenkast:
Well, I don't have a device like that so I can't really test if it works for that dimensions too.. The screenshot I sent is from a 375x667 iPhone 7.

igitur:
image

Is this using that npm package? What's the URL so that I can test it on 320x468?

stenkast:
Yes, it's with the npm package, and it's [this](https://gleeful-dieffenbachia-09d61f.netlify.app/)

igitur:
You use Chrome's normal dev tools and use the responsive layout and set the dimensions to 375x468. Notice that the keyboard is cropped at such a small size. So this PR tries to fix that.
image

stenkast:
I find it better to actually visit the site on the device you're actually testing. Because if you put the dimensions as an iPhone 7, its not going to behave like the url bar is at the bottom. So it's not entirely accurate. 

Fixing that won't account for the Safari problem though, so either way the npm package or a different solution will be required.

Maybe a discord server is in order, to be able to chat about the project more freely.

igitur:
> Fixing that won't account for the Safari problem though, so either way the npm package or a different solution will be required.

Correct. And for that I'm considering both the npm package as well as what @TheKnowles did in https://github.com/theknowles/react-wordle/tree/vh-for-vercel .

@stenkast Would you mind logging a PR that implements the [npm package](https://www.npmjs.com/package/react-div-100vh)?

stenkast:
I think after you've figured this part out, it's easier for you to just implement the npm package yourself, so that it can be tested together with the changes that you have to make to support such small screens.

It also might be preferred to use the other method, to not have to rely on a npm package if it can be avoided and achieve the same consistent result.

igitur:
Dude, I tried the package and didn't get the expected result. Please help out here and submit a PR so that I can see how it works.

stenkast:
Added a PR, my first one so hopefully something didn't go terribly wrong.

PR #495, opened 2022-08-11
Migrate from CRA to Vite

max-programming:
## Migrate from Create React App to Vite

[Vite](https://vitejs.dev) is a great alternative to Create React App. It makes the development workflow a lot faster than CRA. This PR converts CRA to Vite

Just try this out by cloning this branch and installing and running the server with `npm run dev`.

### Remaining stuff

- Move from Jest to [Vitest](https://vitest.dev/)
- Make if statements work in `index.html` like CRA


igitur:
Based on the number of downstream implementations that are dependent on this repo, I think it's unlikely that we can accommodate such an integral framework change.

@cwackerfuss should chime in here.

max-programming:
Did you decide anything about it @cwackerfuss ?

igitur:
Sorry, I don't think this is practical due to reasons mentioned above.

max-programming:
> Sorry, I don't think this is practical due to reasons mentioned above.

I can understand. No probs

PR #498, opened 2022-08-13
Fix prettier plugin incompatibility issue.

igitur:
Because of https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/31 the 2 plugins are incompatible by default. 

The workaround in https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/31#issuecomment-1195411734 solves this issue until it's fixed upstream.


igitur:
Not sure why, but the CI is failing. Issue logged at https://github.com/creyD/prettier_action/issues/95

PR #508, opened 2022-08-21
safari height fix + centered grid

stenkast:
This uses an older sizing method by @igitur at #488 to shrink the keys and cells, together with a npm package to calculate the height, to prevent the url box at the bottom from covering part of the keyboard.

Decided to also center the grid, as I think that's something that should be added.


igitur:
Thanks @stenkast.

So (as you say) this is actually a combination of 
1. adding the npm package
2. incorporating the responsive height changes that I made

The combination of these works on my device. Using only 1. doesn't work on my device, so both are definitely required.

I rebased this PR and split the changes up into the 2 commits to reflect all of this.

Notice that I'm still working on #488 and that there have been changes since you merged it. But if I combine 1. with the latest #488, then I don't get the desired result, so some of my latest changes caused a regression.

I think we are close and that we should use the npm package. I just need to sort out #488 and test on an iPhone 6.

stenkast:
Yeah, I

stenkast:
Whops, wrong button.... 

Yeah, I checked your pull request and saw that too. I'm not entirely sure about what that is about either.

stenkast:
@igitur does the latest preview fix it? I'm using your values and it appears to be working for me on my iPhone 7. Is the navbar auto height necessary? It looks pretty bad.

I added media query to the div surrounding the grid.

stenkast:
This PR can be deleted, as I made the mistake of using my "main" branch for it.. My bad, let me know if you want me to make a new one with the same changes or not.

PR #514, opened 2022-08-25
react-div-100vh + responsive content

igitur:

PR #515, opened 2022-08-25
Removes modal scroll where not needed and modal centering

stenkast:
This removes the modal scroll when it's not needed, and also makes it so the modal respects the height, especially on Safari. Also changes the way it's centering to remove the need for the trick div, if it's required I'll put it back.


stenkast:
Brain is working a bit faster than I can handle. I just realized this is going to need the changes made in #514 first. So I'll close it for now and pick it up later.

igitur:
After #488 and #516 , I don't see scrollbars. Is this PR still necessary?

igitur:
Can you please post before and after screenshots to illustrate the problem and solution?

stenkast:
https://i.imgur.com/Z67Sx6t.png

Left is before, Right is after. It's not really possible to show the scrolling, that's something you'll just have to check yourself.

igitur:
I can see the issue in your screenshot, but I fail to reproduce it on any device. Do you think this is, again, something specific to iOS/Safari? Would you mind confirming that you cannot reproduce it on Chrome/Android on https://reactle.vercel.app/ ?

igitur:
posted under correct PR ;-)

stenkast:
Yeah, it seems to be a Safari issue again, yikes Apple.

igitur (approved):

igitur:
Thanks!

PR #517, opened 2022-08-25
App and related files in its own folder in components + organized test files

stenkast:
I got a lot of ideas for changes to make for organizing but figured a lot of it might be subjective and so I'll just make this one as I think we all can agree on this one.

Changed App to not be a default export, and also moved it and it's related files to its own folder in components for a more cleaner folder structure.

I also made a tests folder in lib, so that they are not mixed together with the main files.


igitur:
I can see the issue in your screenshot, but I fail to reproduce it on any device. Do you think this is, again, something specific to iOS/Safari? Would you mind confirming that you cannot reproduce it on Chrome/Android on https://reactle.vercel.app/ ?

stenkast:
I think you replied to the wrong pull request :) Confirmed, seems to be yet another Safari problem, yikes Apple.

stenkast:
Fair enough

PR #518, opened 2022-08-26
Animation on tiles with status of absent/present/correct only in Info Modal

stenkast:
This is another small quality of life improvement. I think the Info Modal is way too busy in terms of animations with all the irrelevant tiles having the fill animation when you open it, when it's not really necessary.

I think this makes the modal less busy, and puts focus on the tiles that needs it.


igitur:
Again, I can't see the difference on the InfoModal between https://reactle.vercel.app/ and https://react-wordle-git-fork-stenkast-remove-info-c-d4cc73-cwackerfuss.vercel.app/ 

ANOTHER iOS/Safari thing?!?!

stenkast:
> Again, I can't see the difference on the InfoModal between https://reactle.vercel.app/ and https://react-wordle-git-fork-stenkast-remove-info-c-d4cc73-cwackerfuss.vercel.app/
> 
> ANOTHER iOS/Safari thing?!?!

Look at the animation of the tiles that does not have a specific status on them.. In the before they have the "fill" animation when you open it, but on the after I've removed that because I think it's too much animation on unimportant tiles (the status ones need focus) and it looks better like this imo. It's also how the original Wordle does their info screen.

igitur:
Ah, got it.

igitur (approved):

PR #520, opened 2022-08-30
Add link to Western Armenian version

sourencho:


igitur:
Thanks for addition, but please add this link in the correct alphabetical order.

PR #567, opened 2022-11-29
Added German Harry Potter wordle to Readme.md

A-wels:
Since this does not really fit into the existing categories, I added a new one. I feel like adding this to the normal 'fun themes' list would make the list worse for the majority of users, while using the 'other languages' list would not fit into the 'default wordle' theme of the list.


igitur:
Agree, but please add the new category directly below the `Fun Themes` category instead of below the `Maths` category.

igitur:
I'm getting HTTP 502 for your site.

A-wels:
Whoops, I was rebuilding the site. It should work now.

A-wels:
The order of categories has also been changed as requested

PR #590, opened 2023-01-11
Wordle in Kazakh version

Marat1Kh:
Wordle in Kazakh version which has more than 3500+ words


igitur:
Not meant for this repo.

PR #591, opened 2023-01-16
Fix Genele typo

EFord36:
Hi,

It looks like there's a typo within the section `## Projects built using this repo` for the `Genele` project, which is given as `Genel`. It looks to me like its `Genele` from [its homepage](https://andrewholding.github.io/gene-wordle/):

Screenshot 2023-01-16 at 11 49 14

and the link to it [here](https://www.genenames.org):

Screenshot 2023-01-16 at 11 49 23

To be explicit, I'm not associated with Genele, I was just looking at Genele and react-wordle and noticed the discrepancy.

Thanks for making react-wordle!


igitur:
Thanks!

PR #684, opened 2023-07-24
Add build for Docker image

tpansino:
## Description

- Adds a GitHub Actions workflow which
  - Builds the Docker image on each merge to the `main` branch (or each time a tag is pushed, in case the project ever starts publishing releases)
  - Tags the image with a `sha-*` tag of the commit SHA, and the `latest` tag if it's the latest commit to `main`
  - Pushes the tagged images to a (free) GHCR image by the same name (for this repo, would be `cwackerfuss/react-wordle`)

## Why?

It's mildly inconvenient for Docker users to have to manually rebuild the Docker image every time there's a change to this repo. Docker images are free to host on GHCR for public GitHub repos like this one, which makes it easy for us to serve an up-to-date image to the community for each tag or merge to `main`.

## Instructions prior to merging

@cwackerfuss, since this repo is hosted on your personal account, you'll need to do the following to set up the GHCR image for the first time:

- In a terminal on a UNIX host running Docker
    - Make a junk image: `echo 'FROM alpine:latest' | docker image build -t ghcr.io/cwackerfuss/react-wordle --file - .`
    - [Authenticate with GHCR](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic)
    - Push junk image to GHCR to get it to initialize a new package: `docker image push ghcr.io/cwackerfuss/react-wordle`
- Log in to GitHub
    - Navigate to the [list of your packages](https://github.com/cwackerfuss?tab=packages). You should see a `react-wordle` package now, click on it.
    - Scroll down and click on the "Connect Repository" button, select the `cwackerfuss/react-wordle` repo and submit. This will give the package the same permissions as the repo, ie: anyone who is a Maintainer or Admin of the repo will have the same sort of permissions on the package, which is generally what you want.
    - On the right pane, click the "Package Settings", scroll down, and change the package visibility to public, so others can download it.
    - Navigate to the [repo settings](https://github.com/tpansino/react-wordle/settings/actions), and make sure the "Workflow permissions" are set to "Read and write permissions". This is necessary so that GH Actions can get permissions to push to GHCR. The default is by design overly permissive, so I restricted the permissions within the new workflow to only those needed. You may wish to do the same for other workflows, in which case [read this](https://github.com/tpansino/react-wordle/settings/actions).
- Merge this PR. Watch the `Build Docker Image` workflow job to make sure it succeeds. If it doesn't... open an Issue on the repo, and @ me 😄, it's probably an issue with permissions that I wasn't able to test...


igitur:
Hi,

I'm not opposed to this in pricincple, but I do want to ask how much value this is, given that this repo is meant to serve as a base for other people to implement their own Worlde spin-offs in. How would the hosted docker images be used in such cases?

tpansino:
Great questions.

> I do want to ask how much value this is, given that this repo is meant to serve as a base for other people to implement their own Worlde spin-offs in

The project looks to me like it can be used as-is, correct? It's not required to fork it to get value, I can just build the Docker image, start hosting and using it, yes? So the value here is removing the step of users having to clone the repo and build it themselves, and also having to re-clone and re-build on every push to `main` to stay up-to-date. As someone who hosts about 40 different projects on my home server via Docker, I've observed that most projects that are Dockerized have an official image build like this.

Another value is that, by having an existing Docker image, if I want to make a small change (eg: I want to add a Docker `HEALTHCHECK` instruction), I can do that by extending the official image in my own Dockerfile (eg: `FROM cwackerfuss/react-wordle:latest`), without having to clone the entire repo and modify the existing Dockerfile. This again eliminates the need for me to regularly `git pull` the repo, deal with any potential merge conflicts, and rebuild. I can instead just run a `docker build --pull` with my local Dockerfile.

> How would the hosted docker images be used in such cases?

I believe it would be used mostly by self-hosters like myself. I believe we're a growing movement of people who are tired of prices increasing for subscription service, and tired of services being shutdown with no replacement offering, and who have the skills necessary to run such services from a home server. If NYT ever shuts down or starts charging for Wordle for example, I believe this project will see a large increase in usage. :)

Docker has greatly lowered the effort necessary to spin up new services and share projects. Providing an official image build lowers the effort further. After this PR, the only thing a user needs to do is a `docker run` command or create a basic `docker-compose.yml` file.

I also just updated the README to point users to the official build as well.

igitur:
> The project looks to me like it can be used as-is, correct? It's not required to fork it to get value, I can just build the Docker image, start hosting and using it, yes?

Yes, you could do that, and you'd have an exact replicate as what's hosted on https://reactle.vercel.app . Again, I don't see why anyone would want to host at a different URL the exact same game with the exact same solutions on the same game dates as to what's already running at https://reactle.vercel.app . Everybody here is taking this repo as the base case for a derivative implementation, either in a different language, or with a different theme.

> I believe it would be used mostly by self-hosters like myself. I believe we're a growing movement of people who are tired of prices increasing for subscription service, and tired of services being shutdown with no replacement offering, and who have the skills necessary to run such services from a home server. If NYT ever shuts down or starts charging for Wordle for example, I believe this project will see a large increase in usage. :)

If that happens and people use this repo, they'd be building on top of this, not just reusing this as is.

I'm now less convinced that the added complexity to support this adds any value.

tpansino:
It sounds like your mind is made up, but I'll respond for posterity.

> Again, I don't see why anyone would want to host the exact same game with the exact same solutions on the same game dates as to what's already running at https://reactle.vercel.app/

For the same reason I avoid using most Google services - it could disappear at _any_ time.

I have family and friends who are interested in playing Wordle in friendly competition, but we don't like the curation that NYT does with their words, so I was looking for a self-hostable version that we could all use together, one that I control, and that won't disappear or start charging us.

This project is listed in the [Awesome Selfhosted list](https://github.com/awesome-selfhosted/awesome-selfhosted#games). In my experience, that usually implies the creators are OK with the project being hosted as-is, or forked. If you believe people should always fork this, then I recommend you mark the repo as a [template repo](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository) in GitHub so people are aware of your intention. I believe that adds a small banner to the top of the GitHub page to show it's a template.

> Everybody here is taking this repo as the base case for a derivative implementation, either in a different language, or with a different theme.

Not me, but I guess I'm alone. 🤷

> If they dockerize the derivative, they'd have to build the docker image themselves.

Yes, but the way I've implemented the workflow, it would build their docker image automatically for them as well once they create and link a GHCR image by the same name as their fork. It sounds trivial to set that up, but I know GitHub Actions well, and it still took me about 2 hours to get all of those permissions figured out myself.

> If that happens and people use this repo, they'd be building on top of this, not just reusing this as is.

Maybe I'm the only person who feels this way then, but I would rather contribute back features and bugfixes to an upstream repo and not have to maintain my own fork for each of the 40 services I run. Too much overhead.

> I'm now less convinced that the added complexity to support this adds any value.

Sorry to have wasted your time then. I'll close the PR.

